PR #140, opened 2023-06-17
front page

dni:
screenshot-1687281435

- reintroduce frontpage


armurbalda:
Really nice! Maybe let's reduce the area that activates glow/swap box to just the swap box. Right now it's the entire width of the page.

armurbalda (changes requested):
By clicking around in header menu I can pretty easily get it into a state where web app gets stuck in one path (/swap) in example below and doesn't move to other paths anymore. Also happened on /

[Screencast from 06-21-2023 01:33:39 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/16877539/4745e1f7-088c-4840-8b5b-31dbd366fc9e)

EDIT: it is caused by this error, once it appears web app doesn't redirect anymore
```
index-d39623ad.js:149 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'reverse')
    at get each [as each] (index-d39623ad.js:149:30995)
    at index-d39623ad.js:1:9328
    at Object.fn (index-d39623ad.js:1:6690)
    at C4 (index-d39623ad.js:1:4144)
    at $0 (index-d39623ad.js:1:4085)
    at er (index-d39623ad.js:1:2194)
    at T4 (index-d39623ad.js:1:9316)
    at index-d39623ad.js:1:7569
    at Mi (index-d39623ad.js:1:2277)
    at Oe (index-d39623ad.js:1:7562)
```

armurbalda:
correct me if I'm wrong but isn't that hardcoded and should rather be in `i18n.js` too?

image

armurbalda (changes requested):
`Our Lightning Node` with amboss link as title for note stats on frontpage, rm from footer

src/Footer.jsx
```diff
@@ -52,7 +52,15 @@ const Footer = () => {
...
-            <p>{t("footer")}</p>
+            <p>
+                {t("footer")}
+                <br />
+                <a
+                    target="_blank"
+                    href="https://amboss.space/node/026165850492521f4ac8abd9bd8088123446d126f648ca35e60f88177dc149ceb2">
+                    {t("lightningnode")}
+                </a>
+            </p>
```
michael1011, line 63: Still don't like that link here

michael1011 (changes requested):

src/Footer.jsx
armurbalda: Me neither, it belongs as title for the node stats on the front page, see https://github.com/BoltzExchange/boltz-web-app/pull/140#pullrequestreview-1491070483

PR #159, opened 2023-06-20
optimize smartphone

dni:
screenshot-1687276566
screenshot-1687276558

i think it nicer without the box for small devices!


armurbalda:
Honestly I like it better *with* box. The dashed input fields and denominations/fee items look a bit out of place without it. Also it distinguishes the "swap box" better from the footer/header and background. Just my 2 cts.

PR #165, opened 2023-06-21
feat: add spaces to format sats amount

### LINKED ISSUE #146: Add thousands separator
User suggested to have thousands separators for sats denomination. Which makes a lot of sense. given that the lowest number is sth like 10000 or 50000

Instead of going for `,` though, which is U.S. Standard, avoiding confusion with the mostly European `.` standard, I'd do what Breez does: spaces

image

dni:
Closes #146

### DIFF
--- a/src/Create.jsx
+++ b/src/Create.jsx
@@ -303,7 +303,10 @@ const Create = () => {
         let keycode = theEvent.keyCode || theEvent.which;
         if (keycode === 37) {
             // left arrow
-            if (input.value[input.selectionStart - 2] === " ") {
+            if (
+                input.value[input.selectionStart - 2] === " " ||
+                input.value[input.selectionStart - 2] === "."
+            ) {
                 input.setSelectionRange(
                     input.selectionStart - 1,
                     input.selectionStart - 1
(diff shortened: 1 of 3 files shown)


armurbalda:
One can still delete spaces. Behavior should be like this:

https://github.com/BoltzExchange/boltz-web-app/assets/16877539/3cf35afc-747b-4b40-9004-6ae799b8f858

armurbalda (changes requested):
One can still delete spaces. Behavior should be like this:

https://github.com/BoltzExchange/boltz-web-app/assets/16877539/3cf35afc-747b-4b40-9004-6ae799b8f858

armurbalda (changes requested):
Pretty solid, only found one thing that I would love to get fixed since I expect many users to do that: when I delete a number, cursor jumps to to position -1 so when I type a new number it's unexpectedly at position -1 instead of original position. 

[Screencast from 07-24-2023 03:08:40 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/16877539/3ab1c75d-7dfb-431c-b35a-358a1728c3b0)

kilrau:
Marina did it

image

michael1011:
> Marina did it

That's just for display purposes. Not for entering amounts. Very different thing

kilrau:
> > Marina did it
> 
> That's just for display purposes. Not for entering amounts. Very different thing

true that

PR #172, opened 2023-06-27
make expected amount clearer

### LINKED ISSUE #167: [Normal Swaps] Show on-chain amount
Show exact on-chain amount to be sent as instruction sentence similar to

image

We still have users sending wrong amount (e.g. ignoring fees) causing swaps to fail. Also some just don't want to use the copy buttons.

dni:
Closes #167 
screenshot-1687854985

### DIFF
--- a/src/css/index.css
+++ b/src/css/index.css
@@ -550,3 +550,18 @@ textarea {
     background-position: right;
     background-size: 17px;
 }
+
+.break-word {
+    word-break: break-all;
+}
+
+.address-box {
+    background-color: #545c65;
+    background-color: rgba(0, 0, 0, 0.2);
+    border-radius: 7px;
+    padding: 10px;
+    border: 1px solid rgba(255, 255, 255, 0.1);
+    color: rgba(255, 255, 255, 0.3);
+    margin-top: 9px;
+    cursor: pointer;
+}
--- a/src/i18n/i18n.js
+++ b/src/i18n/i18n.js
@@ -52,7 +52,7 @@ const dict = {
         pay_swap_404: "Swap not found!",
         pay_timeout_blockheight: "Timeout blockheight",
         pay_expected_amount: "Expected amount",
-        send_to: "Send {{ amount }} {{ denomination }} to address",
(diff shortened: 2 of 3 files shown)


armurbalda:
Can we put this in a natural sentence as instruction **on top** of the QR and add the estimated time for the timeout blockheight in localized timezone in 24h format too? `getTimezoneOffset();` should give you the UTC+-X offset.

`Please send exactly 1005340 [sats symbol] before timeout blockheight 315 (~16:24 UTC+5)`

Other feedback we got for this page will be in separate issues.

armurbalda (changes requested):
Can we put this in a natural sentence as instruction **on top** of the QR and add the estimated time for the timeout blockheight in localized timezone in 24h format too? `getTimezoneOffset();` should give you the UTC+-X offset.

`Please send exactly 1005340 [sats symbol] to the address below before timeout blockheight 315 (~16:24 UTC+5)`

Other feedback we got for this page will be in separate issues.

dni:
screenshot-1687862533
i like both top and bottom

armurbalda:
> screenshot-1687862533 i like both top and bottom

That's dupblicate information and the space below the QR code will be needed to actually show the address which is other user feedback we got. But as I said we'll do this separately. Also please pay attention to details like sats symbol instead of `sat`. We shouldn't have `sat` or `btc` anywhere, only symbols.

armurbalda:
More user feedback we recently received asked to display the deposit address in this page with the goal to visually check copied address equals address shown in web app, which makes a lot of sense. So I'd say let's try what I suggested before:
image

- `Please send exactly 10157 [sat symbol]` in big letters above QR
- address below QR
- also: let's swap `Copy address` with `Copy amount` button. Copy Address button is most used, hence on mobile right lower area is most accessible also it's common UX to have the copy button on the right on desktop.

armurbalda:
This is how green does it:
image

dni:
screenshot-1689063672

armurbalda (approved):
LGTM

image

PR #184, opened 2023-06-28
feat: Japanese translation

armurbalda:


armurbalda:
Waiting for Koji's review

dni (approved):

armurbalda:
Thank you so much for finalizing Japanese! @yuyaogawa 

I'll add Spanish. You German @dni ?

The language selector itself needs some adjustments:
- [ ] When I am in a non-English language and click to open the selection it already changed to English. It should require to clicks. I suggest we show the "[Globe Icon] Language" when menu is folded in. It's standard.
- [ ] All languages should appear in their native form: English, 日本語, Deutsch, Español in the dropdown.
- [ ] Not sure if this is easy but it's not great that the language selector position jumps because e.g. Japanese needs less space then other languages. Is it possible to assign a fixed min width for header menu items? @dni If a language needs *more* space then the header menu still moves, but at least it's limited.

PR #200, opened 2023-07-18
add warning to claim the swap

### LINKED ISSUE #189: [Reverse Swaps] "Return to this page after paying invoice" Message
Even though we moved to [0-conf claiming](https://github.com/BoltzExchange/boltz-web-app/issues/92), user's on mobile still tend to take some time to pay invoice, browser gets killed by OS and they simply never go back to the browser page waiting in their wallet for funds to arrrive.
image

So we need a "Important: Return to this page after paying invoice" message on this page:
image

dni:
closes #189 

screenshot-1689694485

Important: Return to this page after paying invoice
or
Important: Return to this page after paying this invoice

### DIFF
--- a/src/i18n/i18n.js
+++ b/src/i18n/i18n.js
@@ -118,6 +118,7 @@ const dict = {
         peers: "Number of Peers",
         num_channels: "Number of Channels",
         beta_caution: "BETA - USE WITH CAUTION!",
+        warning_return: "Important: Return to this page after paying invoice",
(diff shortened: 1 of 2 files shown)


armurbalda:
`Important: Return to this page after paying invoice`

Also it should be on top of QR to be seen, timeout height can be below.

armurbalda (approved):
LGTM

PR #216, opened 2023-07-24
fix: improve language selector

### LINKED ISSUE #214: Language selector fixes
The [language selector](https://github.com/BoltzExchange/boltz-web-app/pull/184) needs some adjustments:
- [x] #217
- [x] https://github.com/BoltzExchange/boltz-web-app/pull/218 made languages appear appear in their native form: English, 日本語, Deutsch, Español in the dropdown. TODO: make it part of localization in `i18n.js`
- [x] default to English if string is missing
- [x] ~Not sure if this is easy but it's not great that the language selector position jumps because e.g. Japanese needs less space then other languages. Is it possible to assign a fixed min width for header menu items? @dni If a language needs *more* space then the header menu still moves, but at least it's limited.~

dni:
closes #214


armurbalda:
Hmm just tried the vercel deployments, doesn't fix any of https://github.com/BoltzExchange/boltz-web-app/issues/214 . What am I doing wrong?

src/i18n/i18n.js
```diff
@@ -254,4 +254,13 @@ const dict = {
     },
 };
 
+Object.keys(dict).map((key) => {
+    if (key == "en") return dict[key];
```
michael1011, line 258:
```suggestion
    if (key == "en") return;
```

PR #224, opened 2023-07-26
feat: improve on language menu selector

dni:


armurbalda (changes requested):
- Overlap doesn't look great. Not sure how to fix it though
image 
- Same on mobile
image
- Why remove the `|` separator? Better with it imo.
image
- language selector should close on click in outer area (always has been, but let's fix it in this PR if not too much)

michael1011:
> Overlap doesn't look great. Not sure how to fix it though

Then we leave it as is

> Why remove the | separator? Better with it imo.

Because that is the separator for tabs and looks better without

PR #227, opened 2023-07-27
add timeout to language dropdown menu

dni:


armurbalda:
Yeah works to be honest. Mostly saves you the clikc in the outer area to close selector. Tested: On mobile it doesn't close, which is how it should be 👍🏽

PR #228, opened 2023-07-27
feat: gitbook

armurbalda:
Ready for review now. Comments here in the source pls.

Preview on Gitbook: https://boltz.gitbook.io/new-docs/v/client-boltz-web-app/


README.md
```diff
@@ -1,3 +1,8 @@
+---
+cover: .gitbook/assets/Screenshot from 2023-07-27 17-22-02.png
+coverY: 0
+---
+
```
michael1011, line 5: That cover image looks real bad. Not our font

kilrau: I don't see how it's real bad. They are the same for docs main, boltz-backend, boltz-lnd and webapp. And it's the font we use for marketing stuff like https://twitter.com/Boltzhq/header_photo.

kilrau (approved):

dni (approved):

Issue #229, opened 2023-07-27
WebLN enable() should not expect an object?


otech47:
According to the spec at webln.guide the [enable](https://www.webln.guide/building-lightning-apps/webln-reference/webln.enable) function has a return type of `void`

boltz.exchange has 2 instances calling `window.webln.enable()` and both are expecting it to return an object with a `enabled` field that must be truthy in order for webLN to proceed and work properly

Seems Alby as a WebLN provider is also deviating from this spec since the webln object injected by Alby has an `enable` function that does not have a `void` return type, which is probably why the logic below was written this way

![Screenshot 2023-07-27 at 1 44 10 PM](https://github.com/BoltzExchange/boltz-web-app/assets/4914611/88ad2486-9c02-4d70-a53b-84dee71a050b)

Not sure if the spec should change or this implementation should change but WebLN providers who implement `enable` according to the spec will not be compatible with boltz.exchange due to this inconsistency

humansinstitute:
Hi All - when I test this in Fedi App I'm still getting failures trying to pay via webLN is this deployed on bolts.exchange?

PR #230, opened 2023-07-27
fix: enable() should not expect an object

otech47:
may fix https://github.com/BoltzExchange/boltz-web-app/issues/229

This implementation should still be compatible with providers that return an object on `enable()` but will also allow WebLN providers that do not return anything when `enable()` is called to reach the call to `makeInvoice` so users can make payments on boltz.exchange even if they are not using the Alby extension

also see https://github.com/getAlby/lightning-browser-extension/issues/2596 for more context on spec considerations

### DIFF
--- a/src/Create.jsx
+++ b/src/Create.jsx
@@ -160,14 +160,16 @@ const Create = () => {
     };
 
     const createWeblnInvoice = async () => {
-        let check_enable = await window.webln.enable();
-        if (check_enable.enabled) {
+        try {
+            await window.webln.enable();
             let amount = Number(receiveAmount());
             const invoice = await window.webln.makeInvoice({ amount: amount });
             validateAmount();
             log.debug("created webln invoice", invoice);
             setInvoice(invoice.paymentRequest);
             validateAddress(invoiceInputRef);
+        } catch (error) {
+            log.error(error);
         }
     };
 
--- a/src/status/SwapCreated.jsx
+++ b/src/status/SwapCreated.jsx
@@ -7,11 +7,13 @@ const SwapCreated = () => {
     const [t] = useI18n();
 
     const payWeblnInvoice = async (pr) => {
-        let check_enable = await window.webln.enable();
-        if (check_enable.enabled) {
+        try {
+            await window.webln.enable();
             const result = await window.webln.sendPayment(pr);
             log.debug("webln payment result:", result);
             fetchSwapStatus(swap());
+        } catch (error) {
+            log.error(error);
         }
     };
 


michael1011 (approved):
Thanks for the contribution!

Do you happen to know a WebLN implementation that does not return an object? [All browser extensions mentioned here](https://www.webln.guide/ressources/webln-providers#desktop-browser-extensions) are unmaintained apart from Alby and Kollider which both return an object

otech47:
@michael1011 thanks for review, will fix prettier issues caught by CI shortly

The Fedi Alpha mobile app has a built-in web browser that uses https://github.com/hsjoberg/react-native-webln and when webln is injected, the `enable()` method does not return anything (see [here](https://github.com/hsjoberg/react-native-webln/blob/master/injection/inject.ts#L44))

any react native mobile apps that uses this library it will have the same problem. seems the maintainer @hsjoberg has also used this library in [Blixt wallet](https://github.com/hsjoberg/blixt-wallet) but other than that I'm not aware of any major WebLN providers and how their `webln` objects behave

In my opinion, running `enable()` to ensure the user provides proper authorization makes some sense for something like a browser extension, but an application using its a return value to determine whether to proceed with other webln functions seems like a weak/flaky authorization step

if anything the webLN provider (Alby, etc) should be responsible for blocking webln function calls by tracking whether `enable()` has been called first, That way applications like boltz don't have to worry about that responsibility at all just call it once and proceed once it has resolved

maybe I'll tag a few people here for feedback 🙏🏼 
@wbobeirne
@getAlby
@bumi

michael1011:
Yes, that makes sense. I don't see any downsides to this approach

hsjoberg:
For what it's worth, `enable()` should not return anything according to the specs.

https://github.com/joule-labs/webln/blob/961f02edf0ba9f99235eb28e29e981acd24158f6/src/provider.ts#L43

```Typescript
export interface WebLNProvider {
  enable(): Promise<void>;
```

fiatjaf:
Where are new implementors and existing implementors supposed to learn about what are the new ideas that break everything?

wbobeirne:
https://webln.dev are the original docs I wrote, I didn't originally document enable in lieu of requestProvider but I should add a page or add enable notes to requestProvider.

The docs the alby team forked off of that and maintain also claim a void response https://www.webln.guide/building-lightning-apps/webln-reference/webln.enable. So I'm not sure when implementations started returning an object.

armurbalda:
Good that now we support both - real world *and* specs. Thank you for your contribution! @otech47

PR #242, opened 2023-08-18
[FEAT] do not assume an amount

### LINKED ISSUE #137: Default to empty input field with cursor active in "send" input field
This is the state we want when page load is done, send amount input field ready to take input, ideally also with a `0` as place holder like UniSwap does:

image

Reason?
- we have virtually no chance of correctly guessing the amount the user wants to swap, we want to save the user trouble having to delete the prefilled amount as first step
- it seeems to be industry standard, see above

dni:
closes #137

### DIFF
--- a/src/Create.jsx
+++ b/src/Create.jsx
@@ -278,9 +278,8 @@ const Create = () => {
         const setCustomValidity = (val, isZero) => {
             [sendAmountRef, receiveAmountRef].forEach((ref) => {
                 ref.setCustomValidity(val);
-                ref.classList.add("invalid");
-                if (isZero || val === "") {
-                    ref.classList.remove("invalid");
+                if (!isZero && val !== "") {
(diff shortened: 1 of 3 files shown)


src/Create.jsx
```diff
@@ -70,20 +70,13 @@ const Create = () => {
...
         if (minimum() === 0) {
             return;
         }
-
-        if (firstLoad() && sendAmount() === BigInt(0)) {
-            setFirstLoad(false);
-            setSendAmount(BigInt(minimum()));
-            setReceiveAmount(BigInt(calculateReceiveAmount(minimum())));
-        }
     });
```
michael1011, line 80: That whole effect can be deleted

michael1011 (changes requested):
image

The default state should not be an error. How about "allowing" 0 amounts in the amount validation but keeping the button inactive?

PR #249, opened 2023-09-15
rm: japanese language

kilrau:


dni:
why do we remove it?
if changes happen DH could just PR here and fetch from upstream.
we could just remove it from our language selector

dni (approved):

Issue #256, opened 2023-09-27
Misleading "insufficient fee, rejecting replacement" error message when swap was successful


kilrau:
Very confusing since the swap was actually successful, but my eyes were on the error message

[Screencast from 09-27-2023 01:40:18 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/b0c58956-1907-48e6-b0cc-9237e3083846)

kilrau:
THIS IS NEW, doesn't happen with current deployment on boltz.exchange

dni:
hm, i cant reproduce it

michael1011:
Still happening

michael1011:
closed by #261

PR #260, opened 2023-09-30
fix: issue claiming in background

dni:
the if failed when swap() is set to null onCleanup()

Closes https://github.com/BoltzExchange/boltz-web-app/issues/256


kilrau:
just tested via https://webapp-mainnet-3fcoxsyka-boltz.vercel.app/ - error is still there

image

PR #261, opened 2023-09-30
fix: extra check if swap was already claimed

### LINKED ISSUE #256: Misleading "insufficient fee, rejecting replacement" error message when swap was successful

dni:
closes #256


kilrau (approved):
#256 error is gone :100:

PR #266, opened 2023-10-02
refactor: use polling to fetch latest status of background swaps

michael1011:


kilrau (changes requested):
Tests need adjustment
image

dni (approved):

kilrau (approved):
Started 13 reverse swaps without paying any invoice
image

Then paid invoice of the first one. Noticed a couple of 404s, but could observe the swapstatus polling working
image
image

Swap was claimed without issues, worked!

image

PR #269, opened 2023-10-03
fix: default swap direction

### LINKED ISSUE #268: Change default pair back to LN->BTC
Boltz now defaults to BTC->LN for new users (test by opening boltz.exchange in a private tab it). Let's revert to LN->BTC as this is still our #1 pair by number of swaps. Fwiw I suspect https://github.com/BoltzExchange/boltz-web-app/pull/257, but I might be wrong.

dni:
closes #268


kilrau (approved):
Works, default is back to LN->BTC

PR #276, opened 2023-10-10
feat: use invoice amount from pasted invoice

### LINKED ISSUE #252: [Normal Swaps] If no amounts entered manually, web app should auto-fill based on invoice
This should autofill amounts in swap boxes
image

If user then changes amounts after the auto-fill, behavior stays as-is: clear invoice field

dni:
closes #252

https://github.com/BoltzExchange/boltz-web-app/assets/1743657/5524c008-4353-40a2-87c4-7e442c4d2755

### DIFF
--- a/src/Create.jsx
+++ b/src/Create.jsx
@@ -125,7 +125,7 @@ const Create = () => {
 
     const navigate = useNavigate();
 
-    const checkInvoice = () => {
+    const resetInvoice = () => {
         if (isInvoice(invoice())) {
             setInvoice("");
             setInvoiceValid(false);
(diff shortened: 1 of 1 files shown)


kilrau (changes requested):
Behavior is good now. The edge case where user did change amounts after pasting invoice can't happen since 0 amounts are now prefilled and and changing amounts  from non-0 after pasting invoice clears the invoice. Good stuff, works as expected!

Not related to this PR, but let's fix it up in this one: when I paste a valid invoice and then delete some char(s) the validatio should immediately deactivate the `Create Atomic Swap` button. Right now it stays active (even though it's not working).

kilrau (approved):
Moved https://github.com/BoltzExchange/boltz-web-app/pull/276#pullrequestreview-1666999510 into a new issue https://github.com/BoltzExchange/boltz-web-app/issues/277 since it's not related to PR and we should get this deployed fast - several users are struggling with this.

PR #283, opened 2023-10-16
fix: check fee amount instead of pair hash

### LINKED ISSUE #197: Hash change causes "network fees were updated" when they were in fact not updated
Especially relevant for Liquid Swaps since fees practically never change there, but notification pops up because mainchain fees changed. Reason: pairhash contains fees for more than relevant assets

Diffuculty: we don't know who is using it so me might break an implementation if we change it in backend. So let's try to do it in frontend.

Another thing to consider speaking *for* a change in backend: e.g. Liquid wallets using the pairhash feature the same way will face the same issue

### LINKED ISSUE #390: Subtle redesign of denomination switch
to look a bit more like a switch like with a slightly different blue switch background and (grey below) and a brigher "active" blue (pink below) to resemble toggle switch design:
image

dni:
closes #197

### DIFF
--- a/src/Create.jsx
+++ b/src/Create.jsx
@@ -215,7 +215,7 @@ const Create = () => {
             };
         }
 
-        if (!await feeCheck(t("feecheck"))) {
+        if (!(await feeCheck(t("feecheck")))) {
             return;
         }
 
--- a/src/Create.jsx
+++ b/src/Create.jsx
@@ -215,6 +215,10 @@ const Create = () => {
             };
         }
 
+        if (!await feeCheck(t("feecheck"))) {
+            return;
+        }
+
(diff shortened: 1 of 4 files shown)


kilrau:
Hmm /createswap gets a 400 
image

[Screencast from 10-16-2023 03:05:03 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/22520a4a-f50d-466c-b4f1-d171ca6f35c6)

kilrau:
With a new invoice now it works

kilrau (approved):
Tested both directions: ln -> liquid, liquid -> ln both waiting until boltz fee changed (had another boltz tab open to confirm mainchain fee changed) and worked fine without error message

michael1011 (approved):

PR #284, opened 2023-10-16
fix: only claim from inside swapchecker

dni:
closes #261

### DIFF
--- a/src/Pay.jsx
+++ b/src/Pay.jsx
@@ -36,9 +36,7 @@ const Pay = () => {
     createEffect(() => {
         let tmpSwaps = swaps();
         if (tmpSwaps) {
-            let currentSwap = tmpSwaps
-                .filter((s) => s.id === params.id)
-                .pop();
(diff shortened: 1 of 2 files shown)


src/Pay.jsx
```diff
@@ -32,20 +34,27 @@ const Pay = () => {
...
-                .filter((s) => s.id === params.id)
-                .pop();
-            if (current_swap) {
-                log.debug("selecting swap", current_swap);
-                setSwap(current_swap);
-                fetchSwapStatus(current_swap);
-                setReverse(current_swap.reverse);
+        let tmpSwaps = swaps();
+        if (tmpSwaps) {
+            let currentSwap = tmpSwaps.filter((s) => s.id === params.id).pop();
```
michael1011, line 39:
```suggestion
            const currentSwap = tmpSwaps.filter((s) => s.id === params.id).pop();
```

michael1011 (approved):
Works!

PR #285, opened 2023-10-16
feat: change last changed send/receive amount on fee change

### LINKED ISSUE #245: Persist amount that was entered manually on fee update
When eg the onchain amount is entered manually for reverse swaps, that amount should stay the same after fee changes. The amount that wasn't entered by the user should be changed

dni:
closes #245

### DIFF
--- a/src/signals.js
+++ b/src/signals.js
@@ -1,7 +1,7 @@
 import { createEffect, createSignal } from "solid-js";
 import { createStorageSignal } from "@solid-primitives/storage";
-import { LN } from "./consts";
 import { isMobile } from "./helper";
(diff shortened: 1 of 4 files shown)


src/signals.js
```diff
@@ -29,6 +29,7 @@ export const [minimum, setMinimum] = createSignal(0);
 export const [maximum, setMaximum] = createSignal(0);
 
 // swaps
+export const [amountChanged, setAmountChanged] = createSignal(null);
```
michael1011, line 32: Why don't we default to the sideSend and avoid the null case completely?

michael1011 (approved):

PR #289, opened 2023-10-16
feat: use swapbox on `/`

### LINKED ISSUE #255: Landing page swap box should have cursor in input field that I clicked on when jumping to /swap
So here I click on the `Enter BTC address` input field, so the cursor should be in this field ready to take input. The same way when I click on the lower receive amount input field it should be there.

Instead cursor always is in upper send asset input field when jumping to /swap.
[Screencast from 09-27-2023 01:32:57 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/1bbcb725-86dc-4afe-990d-767d50fdadbc)

dni:
closes #255

not sure if this is a good solution, but the only one i can think of, because routing from `/` to `/swap` always rerenders the Create component, thus loosing the state where you clicked.

### DIFF
--- a/src/Hero.jsx
+++ b/src/Hero.jsx
@@ -41,7 +41,7 @@ const Hero = () => {
         <div id="hero" class="inner-wrap">
             <div
                 id="create-overlay"
-                class={hideHero() ? "glow" : ""}
+                class={hideHero() ? "" : "glow"}
                 onclick={() => setHideHero(true)}>
                 <Create />
             </div>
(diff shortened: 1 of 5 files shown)


kilrau (changes requested):
Almost there, only thing missing is to bring back landing page when logo is clicked

dni:
done

michael1011 (changes requested):
Clicking on the box should change the url to `/swap`. Looks good apart from that

michael1011 (approved):

kilrau (approved):
LGTM, good stuff @dni

PR #290, opened 2023-10-16
Dependency Updates

michael1011:
Updates:

- `prettier`: trailing comma enabled by default now
- `@solid-primitives/storage`: `createStorageSignal` is deprecated and has been refactored to `makePersisted(createSignal` with a custom serializer to support values written by the deprecated function
```js
// To support the values created by the deprecated "createStorageSignal"
const stringSerializer = {
    serialize: (value) => value,
    deserialize: (value) => value,
};
```
- `@solid-primitives/i18n`: has breaking API changes; those have been addressed

The only updates I did not do are:
- `vitest`: breaks our tests completely
- `@noble/secp256k1` because there is no wrapper for the new version to make it compatible with bitcoinjs-lib


kilrau:
> * `@noble/secp256k1` because there is no wrapper for the new version to make it compatible with bitcoinjs-lib

Do you expect this wrapper to be there at some point? If not, let's ping the maintainer

michael1011:
`@noble/secp256k1` was never compatible with `bitcoinjs-lib`, we had our own compatibility wrapper. I switched to this wrapper library that is maintained and tested: https://github.com/bitcoinerlab/secp256k1

PR #303, opened 2023-11-13
feat: make min and max clickable

dni:
screenshot-1699912469
screenshot-1699912564


kilrau (approved):
LGTM

Initially the strong yellow turned me off, but it's growing on and in line with how WEBLN button behaves.

PR #309, opened 2023-11-16
fix: issue 281, Remove timeout blockheight from swap flow

### LINKED ISSUE #281: [UX Revamp] Remove timeout blockheight from swap flow
WILL PROVIDE WIREFRAME

Remove refund info from swap flow. It's only useful when a refund is actually needed -> only in boltz.exhcange/refund
image

dni:
closes #281

before:
screenshot-1700129449

after:
screenshot-1700307267


kilrau:
Can we:
- adjust `btc` denomination to `L-BTC` or `RBTC` or `BTC` depending on swap type
- move `Send 0.001X ...` on top of QR Code
- below address box `Make sure your transaction confirms within {invoice expiry|swapexpiry} hours!

kilrau (approved):

kilrau (changes requested):
Normal Swap shoulds say `sats` if denomination is sats, L-BTC if denomination is full BTC
image

Reverse Swap should say `sats` (plural), L-BTC if denomination is full BTC
image

src/status/SwapCreated.jsx
```diff
@@ -15,42 +15,32 @@ const SwapCreated = () => {
...
             <hr />
+            <h3>{t("warning_return")}</h3>
+            <hr />
             <Show when={webln()}>
                 <span
                     class="btn btn-light"
                     onClick={() => payWeblnInvoice(swap().invoice)}>
                     {t("pay_invoice_webln")}
                 </span>
+                <hr />
```
michael1011, line 43: I don't like this hr

michael1011 (changes requested):
The sats denomination should say sats

@kilrau check the updated i18n strings

src/helper.js
```diff
@@ -50,6 +50,13 @@ const parseBlindingKey = (swap) => {
     return swap.blindingKey ? Buffer.from(swap.blindingKey, "hex") : undefined;
 };
 
+export const cropString = (str) => {
+    if (str.length < 40) {
+        return str;
+    }
+    return str.substring(0, 20) + "..." + str.substring(str.length - 20);
```
michael1011, line 57:
```suggestion
    return str.substring(0, 19) + "..." + str.substring(str.length - 19);
```

Else it won't fit in one line

image

dni:
should now say sats, and also the according assets if denom btc and not an lightingpayment

michael1011 (approved):

PR #310, opened 2023-11-16
feat: alignment past swaps

dni:
closes #291

PR #315, opened 2023-11-16
hotfix: alignment past swaps

dni:
introduced by #310


src/style/history.scss
```diff
@@ -1,6 +1,6 @@
 #past-swaps {
```
michael1011, line 1: Also, this is in the wrong file. The id is defined in `SwapList` which we also use in `Refund`

michael1011 (approved):

PR #317, opened 2023-11-16
refactor: extract create button

dni:
with `create()` action


michael1011 (changes requested):
Clicking create throws: `Uncaught (in promise) ReferenceError: log is not defined`

But concept ack

src/signals.js
```diff
@@ -8,6 +8,7 @@ import { isMobile } from "./helper";
 const defaultSelection = Object.keys(pairs)[0].split("/")[0];
 
 // ui
+export const [buttonLabel, setButtonLabel] = createSignal("");
```
michael1011, line 11: Let's not introduce even more global signals. They are only used in `Create` and its children, so let's declare them in `Create` and pass them to the relevant children as props

src/signals.js
```diff
@@ -36,6 +37,7 @@ export const [receiveAmountFormatted, setReceiveAmountFormatted] =
 export const [refundAddress, setRefundAddress] = createSignal(null);
 export const [onchainAddress, setOnchainAddress] = createSignal("");
 export const [invoice, setInvoice] = createSignal("");
+export const [lnurl, setLnurl] = createSignal(false);
```
michael1011, line 40: Same here

src/signals.js
```diff
@@ -101,6 +103,7 @@ export const [assetSend, setAssetSend] = makePersisted(createSignal(LN), {
 export const [valid, setValid] = createSignal(false);
 export const [invoiceValid, setInvoiceValid] = createSignal(false);
 export const [addressValid, setAddressValid] = createSignal(false);
+export const [sendAmountValid, setSendAmountValid] = createSignal(true);
```
michael1011, line 106: And here

src/Create.jsx
```diff
@@ -329,82 +198,65 @@ const Create = () => {
...
                 input.classList.remove("invalid");
-                setAddressValid(true);
-                setOnchainAddress(inputValue);
             } catch (e) {
-                setAddressValid(false);
-                input.setCustomValidity("invalid address");
-                input.classList.add("invalid");
+                invalidateAddress(
+                    input,
+                    "Invalid onchain address",
```
michael1011, line 235: Translation missing

src/Create.jsx
```diff
@@ -329,82 +198,65 @@ const Create = () => {
...
-                        setInvoiceValid(false);
-                        input.setCustomValidity(e);
-                        input.classList.add("invalid");
-                        return;
-                    }
+            const input = invoiceInputRef;
+            const inputValue = input.value.trim();
+            try {
+                if (isLnurl(inputValue)) {
+                    setButtonLabel(t("fetch_lnurl"));
```
michael1011, line 244: I don't like this. The fetching is already abstracted away, there is no need to inform the user about that explicitly

src/i18n/i18n.js
```diff
@@ -208,6 +213,8 @@ const dict = {
         download_refund_file: "Rückerstattungsdatei herunterladen",
         invalid_refund_file: "Ungültige Rückerstattungsdatei",
         invalid_backup_file: "Ungültige Backupdatei",
+        invalid_invoice: "Not a Lightning invoice, LNAddress or LNURL",
+        invalid_0_amount: "0 amount invoices are not allowed",
```
michael1011, line 217: Not translated

src/i18n/i18n.js
```diff
@@ -351,6 +360,8 @@ const dict = {
         download_refund_file: "Descargar archivo de reembolso",
         invalid_refund_file: "Archivo de reembolso no válido",
         invalid_backup_file: "Archivo de backup no válido",
+        invalid_invoice: "Not a Lightning invoice, LNAddress or LNURL",
+        invalid_0_amount: "0 amount invoices are not allowed",
```
michael1011, line 364: Not translated

src/i18n/i18n.js
```diff
@@ -420,6 +431,7 @@ const dict = {
         not_found: "404 - Página no encontrada",
         not_found_subline: "La página buscada no existe.",
         back_to_home: "Volver al inicio",
+        fetch_lnurl: "Fetch invoice from LNURL / LNAdress",
```
michael1011, line 434: Not translated

michael1011 (changes requested):

src/i18n/i18n.js
```diff
@@ -43,6 +43,7 @@ const dict = {
         footer: "Made with ❤️ by Team Boltz",
         create_channel: "Create Lightning Channel",
         create_channel_subline: "Channel Inbound or Outbound Capacity",
+        create_reverse_swap: "Create Reverse Atomic Swap",
```
kilrau, line 46: NO. Let's not surface normal/reverse swap in the UI.

kilrau (changes requested):
- [ ] When amounts are entered, it should refresh. Needs another string I think: `Please enter destination address` 
Screenshot from 2023-11-20 15-32-55
- [ ] This is broken now, activates `Create Atomic Swap` button when no invoice is supplied. Should instead have `Please enter lightning invoice` instead.
- [ ] Think we should have some tests for things like above not to break down the road: if amount out of min/max range or address/invoice missing -> test that `Create Atomic Swap` button stays deactivated.

Give this another round of fixing and request review when you think it's ready @dni

src/i18n/i18n.js
```diff
@@ -229,6 +233,8 @@ const dict = {
         download_refund_file: "Rückerstattungsdatei herunterladen",
         invalid_refund_file: "Ungültige Rückerstattungsdatei",
         invalid_backup_file: "Ungültige Backupdatei",
+        invalid_invoice: "Not a valid Lightning invoice, LNAddress or LNURL",
+        invalid_0_amount: "0 amount invoices are not allowed",
```
michael1011, line 237: Translations to be done by @kilrau

michael1011 (approved):

PR #318, opened 2023-11-16
refactor: show tx links for RSK

michael1011:
It makes no sense to show the contract address for RSK swaps, because every RSK swap uses the same contract address.

So this PR refactors what we show on the block explorer link for RSK swaps, which is either the lockup or claim transaction, whichever is relevant at a given step

### DIFF
--- a/src/Pay.jsx
+++ b/src/Pay.jsx
@@ -71,7 +71,7 @@ const Pay = () => {
     createEffect(() => {
         const tx = swapStatusTransaction();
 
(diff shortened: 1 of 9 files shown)


kilrau (changes requested):
Makes a lot of sense.
Normal swap: all good

Reverse Swap:
Button shows `lockup transaction` with correct link, but then when claiming showing `claim transaction` the linked tx is still the lockup transaction. Only in the final success screen `Open claim transaction` it links to the actual claim transaction.

PR #325, opened 2023-11-19
feat: refine design homepage

dni:


kilrau:
Had a quick look:
- I really don't wanna go back to the half-visible swap box. Give it a "move" animation on hover to signify it's there to be clicked and used if you want to fancy up the box
- font is very nice, works in FF & Chromium & FF&Vandadium on Android
- centric headlines are nice, not sure what you had in mind to put as description but we could link to both of our nodes, including [CLN](https://amboss.space/node/02d96eadea3d780104449aca5c93461ce67c1564e2e1d73225fa67dd3b997a6018) even though nodestats are [currently](https://github.com/BoltzExchange/boltz-backend/issues/386#issuecomment-1752142975) only LND

kilrau (changes requested):
Generally good, I just don't like this: the swap box being blurred. We got this feedback from users in the beginning: It's not 100% clear if the swap box is active and can be used and actually makes me click randomly into the box trying to make it become active. Take away the blur and instead brighten the glow a bit on hover to have a similar effect. Apart from that font works in FF & Chromium on my linux machine and android phone.

[Screencast from 12-14-2023 12:47:46 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/0a00efcc-35d4-433c-b1a4-fa9feb2a3b1b)

Section titles centered much better but I liked the old style better, titles should be somewhat bright and shiny, now I almost overlook them:
image
vs
image

kilrau (approved):
glow effect almost not visible, but fine with me

src/Hero.jsx
```diff
@@ -79,12 +79,10 @@ export const Hero = () => {
                         <img src={liquid} alt="Liquid Bitcoin" />
                     </div>
                 </div>
+                <h2 class="headline" onclick={openNodeInfo}>
+                    {t("node")}
+                </h2>
```
michael1011, line 84: Cursor is the one for text selection and not a pointer

michael1011 (changes requested):
as @kilrau said, the head lines have no contrast

michael1011 (approved):

PR #329, opened 2023-11-20
fix: handle js error when asset is missing on api

dni:
and default to BTC/BTC. this only happens while development, running on the mainnet api which is still missing the new asset beeing worked on.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -14,12 +14,13 @@ import {
     setup,
 } from "./compat";
 import { pairs } from "./config";
-import { RBTC } from "./consts";
+import { BTC, RBTC } from "./consts";
 import { ECPair } from "./ecpair/ecpair";
 import {
     asset,
     ref,
     refundAddress,
+    setAsset,
     setConfig,
     setFailureReason,
     setNotification,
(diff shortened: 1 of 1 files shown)


src/helper.js
```diff
@@ -100,7 +101,12 @@ export const errorHandler = (error) => {
 };
 
 export const getApiUrl = (asset) => {
-    return pairs[`${asset}/BTC`].apiUrl;
+    const pair = pairs[`${asset}/BTC`];
+    if (!pair) {
+        log.error("no pair found for asset", asset, "using", BTC);
+        return pairs[`${BTC}/BTC`].apiUrl;
+    }
```
michael1011, line 108: This is a bad idea. We should overwrite the asset with `setAsset`, not put a hack in this function

src/helper.js
```diff
@@ -100,7 +101,12 @@ export const errorHandler = (error) => {
 };
 
 export const getApiUrl = (asset) => {
-    return pairs[`${asset}/BTC`].apiUrl;
+    const pair = pairs[`${asset}/BTC`];
+    if (!pair) {
+        log.error("no pair found for asset", asset, "using", BTC);
```
michael1011, line 106: Why `using`?

michael1011 (changes requested):

src/helper.js
kilrau, line 106:
```suggestion
        log.error("API doesn't support any pair for asset", asset, "reverting to", BTC);
```

kilrau (changes requested):
translations missing

PR #337, opened 2023-11-25
added use-case section to landing page

lenekin:
The translations probably need a look over.

### DIFF
--- a/src/i18n/i18n.js
+++ b/src/i18n/i18n.js
@@ -134,6 +134,12 @@ const dict = {
         not_found: "404 - Page Not Found",
         not_found_subline: "The page you are looking for does not exist.",
         back_to_home: "Back to Home",
+        use_case: "Use Cases",
+        reliably_cheap_ln_channel_rebalancing: "Reliably cheap LN channel rebalancing",
+        what_you_do_with_boltz: "What you can do with Boltz",
+        dcaing_from_ln_into_cold_storage: "DCAing from LN into cold storage",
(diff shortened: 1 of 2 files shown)


kilrau (changes requested):
- [ ] Conflicts need to be resolved
- [ ] Doesn't look fitting yet visually since content of this box is taking much more space than the other three. Let's shorten Text as per my suggestions. Please play around with it until it looks balanced with the other boxes. Remove emojis if that makes bullet points fit into one line.

src/i18n/i18n.js
```diff
@@ -134,6 +134,12 @@ const dict = {
         not_found: "404 - Page Not Found",
         not_found_subline: "The page you are looking for does not exist.",
         back_to_home: "Back to Home",
+        use_case: "Use Cases",
+        reliably_cheap_ln_channel_rebalancing: "Reliably cheap LN channel rebalancing",
+        what_you_do_with_boltz: "What you can do with Boltz",
```
kilrau, line 139:
```suggestion
        what_you_do_with_boltz: "Why use Boltz",
```

src/i18n/i18n.js
```diff
@@ -134,6 +134,12 @@ const dict = {
...
+        dcaing_from_ln_into_cold_storage: "DCAing from LN into cold storage",
```
kilrau, line 140:
```suggestion
        dcaing_from_ln_into_cold_storage: "DCA from LN -> cold storage",
```

PR #355, opened 2023-12-11
fix: metamask not installed

dni:
screenshot-1702290134


src/i18n/i18n.ts
```diff
@@ -385,6 +387,7 @@ const dict = {
         send_to: "Enviar {{ amount }} {{ denomination }} a",
         pay_invoice_to: "Pague esta factura de {{ amount }} {{ denomination }}",
         pay_address: "Dirección",
+        no_metamask: "MetaMask not installed",
```
michael1011, line 390: @kilrau Spanish translation pls

michael1011 (approved):

src/i18n/i18n.ts
dni, line 390:
```suggestion
        no_metamask: "MetaMask no instalado",
```

PR #364, opened 2023-12-13
fix: translate create button label on change

### LINKED ISSUE #363: Create button localization doesn't change if I switch language in an min/max amount error state
If swap box shows a min/max amount error because I entered an amount out of bounds and then I go change the language, the swap button language is not changed. For other errors like "Connect Metamask" this didn't happen but didn't test all possible errors.

[Screencast from 12-13-2023 10:17:15 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/c81c496b-a85c-42c2-aeb0-34047e711be2)

dni:
closes #363

### DIFF
--- a/src/components/CreateButton.tsx
+++ b/src/components/CreateButton.tsx
@@ -12,8 +12,11 @@ import t from "../i18n";
 import {
     asset,
     config,
+    denomination,
     invoice,
     lnurl,
+    maximum,
+    minimum,
     onchainAddress,
     online,
     receiveAmount,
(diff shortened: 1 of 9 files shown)


src/components/CreateButton.tsx
```diff
@@ -211,13 +215,32 @@ export const CreateButton = () => {
...
+                amount: formatAmount(maximum()),
+                denomination: denomination(),
+            };
+        } else if (buttonLabel() === "invalid_invoice") {
+            return {
+                asset: asset(),
+            };
+        }
+        return {};
+    };
```
michael1011, line 235: Do not do this here. Label and arguments should be set together

michael1011 (changes requested):

kilrau (changes requested):
Still found some cases like this one where language is not switched:
image

kilrau (approved):

Issue #344, opened 2023-12-03
Add missing translations


kilrau:
Translations to be done by @kilrau

_Originally posted by @michael1011 in https://github.com/BoltzExchange/boltz-web-app/pull/317#discussion_r1413169807_

kilrau:
Will do these last thing before we test a RC

PR #369, opened 2023-12-14
fix: assets depends on config settings

dni:
now asset select will actually use assets configured on config.js

also closes https://github.com/BoltzExchange/boltz-web-app/issues/344

### DIFF
--- a/src/components/AssetSelect.jsx
+++ b/src/components/AssetSelect.jsx
@@ -13,6 +13,9 @@ import {
     setAssetSend,
 } from "../signals";
 
+const assets = Object.keys(pairs).map((pair) => pair.split("/")[0]);
+assets.push(LN);
+
 const SelectAsset = () => {
     const setSelectAsset = (isSend, asset) => {
         const setter = isSend ? setAssetSend : setAssetReceive;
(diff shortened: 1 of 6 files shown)


kilrau (changes requested):
Works on mainnet, on testnet it still happens (see url, it's the testnet deployment from this branch)

image

src/components/AssetSelect.jsx
```diff
@@ -49,6 +50,14 @@ const SelectAsset = () => {
         );
     };
 
+    const assets = () => {
+        const names = Object.keys(pairs).map((pair) => {
+            return pair.split("/")[0];
+        });
+        names.push(LN);
+        return names;
+    };
```
michael1011, line 59: Why is this a function? It does only ever need to be evaluated once. Configs don't do reactive changes.

michael1011, line 59: Why is this a function? It does only ever need to be evaluated once. Configs don't do reactive changes.

michael1011 (changes requested):

michael1011 (approved):

kilrau (changes requested):
Can you rebase this on master so the `C.json` fix is in and I can give this another proper test? @dni

dni:
rebased

michael1011 (approved):

Issue #370, opened 2023-12-14
Getting `C.json` error when rootstock asset was selected before


kilrau:
Getting this `C.json` error in both deployments testnet (!) & mainnet when rootstock asset was selected before

[Screencast from 12-14-2023 01:03:56 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/6ccd0e54-ce05-4049-82db-1a640180a8a0)

_Originally posted by @kilrau in https://github.com/BoltzExchange/boltz-web-app/pull/367#pullrequestreview-1781671783_

PR #371, opened 2023-12-14
fix: only set onchain addr on reverse swap

dni:
Closes https://github.com/BoltzExchange/boltz-web-app/issues/370

also check if defined before extracting


michael1011 (approved):

kilrau (approved):
Works now where it failed before. No `C.json` errors anymore. That's how far my testing can go :)

PR #378, opened 2023-12-20
feat: deeplinks for wallets

### LINKED ISSUE #299: Mobile: Deep link lightning and chain wallets
Tests with Blockstream's [Green](https://blockstream.com/green/) Wallet for Liquid & Chain, Acinq's [Phoenix](https://phoenix.acinq.co/) for Lightning on **Android and iOS** both should pass.

For some sample deep links check:
- https://brqgoo.medium.com/deep-linking-crypto-de7fe10558c5
- https://blog.simple-bitcoin.app/deeplinks-lightning-how-it-works-and-why-its-important-for-bitcoin-ux/

Improves UX on mobile a lot.

EDIT: include https://bitcoinqr.dev/ as valid deeplink with a ln invoice

dni:
add links on SwapCreated and InvoiceSet on the QR code and as button.

its working for alby on browser for me i did not test bip21 yet.

closes #299

screenshot-1703075559

i replaced the copy bip21 for the open in wallet button, usually you don't really want to copy bip21 right? else we get a small ui issue because we have to less room to make it look good.

screenshot-1704406281

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -160,6 +160,7 @@ const dict = {
         scan_qr_code: "Scan QR Code",
         version: "Version",
         commithash: "Commit Hash",
+        open_in_wallet: "Open in Wallet",
     },
     de: {
         language: "Deutsch",
@@ -329,6 +330,7 @@ const dict = {
         scan_qr_code: "QR Code scannen",
         version: "Version",
         commithash: "Commit Hash",
+        open_in_wallet: "Im Wallet öffnen",
     },
     es: {
         language: "Español",
@@ -497,6 +499,7 @@ const dict = {
         scan_qr_code: "Escanear código QR",
         version: "Versión",
         commithash: "Commit Hash",
+        open_in_wallet: "Abrir en monedero",
     },
     zh: {
         language: "中文",
@@ -651,6 +654,7 @@ const dict = {
         scan_qr_code: "扫描 QR 码",
         version: "版本",
         commithash: "提交哈希",
+        open_in_wallet: "在钱包中打开",
     },
 };
 
--- a/src/pages/Pay.tsx
+++ b/src/pages/Pay.tsx
@@ -163,6 +163,7 @@ const Pay = () => {
                         when={
                             swap().asset !== RBTC &&
                             swapStatus() !== null &&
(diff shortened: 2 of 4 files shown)


kilrau (changes requested):
Desktop:
- [ ] For Reverse Swaps on I don't think it makes sense to have the "Open in Wallet" button. It opens Alby, which already has the `Pay invoice via WebLN` button. https://github.com/LN-Zap/zap-desktop is archived, I don't think there is any other popular desktop lightning wallets that could make use of that button.
- [ ] For Normal Swaps, on my ubuntu system the handler couldn't find my Green wallet to open `liquidnetwork:`, I openen an [issue on green](https://github.com/Blockstream/green_qt/issues/103) side. For bitcoin mainchain the handler listed others
image but then send dialogue and amounts couldn't be prefilled.

Summary: not very useful on Desktop, if you insist let's keep it for normal swaps only and we hope that Green & other wallets improve in future. I am also fine removing it completely for desktop.

Mobile:
Works beautifully in all wallets I tested. Everything prefilled as it should be, even in Green.

kilrau (changes requested):
Reverse Swaps:
- [ ] So now `Pay invoice via WebLN` only applies to Desktop, `Open in Wallet` only applies for Mobile. I think we should unify style! Blue button on top vs both yellow next to each other. Don't care which, but it should be the same. `Scan QR Code` on mobile is blue on top, same style as `Pay invoice via WebLN` on Desktop.
image
Screenshot from 2024-01-10 17-19-55

Normal Swaps:
- [x] Mobile is missing `Copy Address` now. Which is only acceptable if we add a copy symbol into the address filed so it's discoverable that the address can be copied by tapping on it. I just really don't like that there are 3 equally looking buttons on mobile and desktop doing different things.
image
image

The "blue button on top" style for `Open in Wallet` can work if you agree doing https://github.com/BoltzExchange/boltz-web-app/issues/444

kilrau:
Since we are now free to do https://github.com/BoltzExchange/boltz-web-app/issues/444, we can use this button style for "open in wallet" on mobile.
image

dni:
closes #444

kilrau (approved):

PR #379, opened 2023-12-20
feat: make btns width 100%

dni:
looks better :)

before/after

screenshot-1703076064
screenshot-1703076055


michael1011 (approved):
imho the buttons should have the same width but ACK

Issue #389, opened 2023-12-21
Create button shows "Invalid L-BTC Address" when amounts are out of bounds


kilrau:
![image](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/f6ac2b85-a4c4-4982-a99b-131f6f21cae5)

PR #393, opened 2023-12-21
fix: revalidate amounts when address is valid

dni:
closes https://github.com/BoltzExchange/boltz-web-app/issues/389


kilrau (approved):
Works as expected now: the error messages updating to the last error that was encountered.

PR #394, opened 2023-12-21
feat: placeholders instead of 0 on init

dni:
screenshot-1703186406


kilrau (commented):
First round of testing - works as expected. Will give it another spin after conflicts are resolved and you consider it final @dni

michael1011 (approved):

Issue #391, opened 2023-12-21
Auto-switch denomination


kilrau:
From btc to sats if amount entered is >99
From sats to btc if amount entered is <1 aka contains `.`

PR #395, opened 2023-12-21
feat: automatic denom switcher

dni:
switches to sat denom of amount > x

closes https://github.com/BoltzExchange/boltz-web-app/issues/391

### DIFF
--- a/src/pages/Create.tsx
+++ b/src/pages/Create.tsx
@@ -21,6 +21,7 @@ import {
 import {
     calculateDigits,
     convertAmount,
+    denominations,
     formatAmount,
     getValidationRegex,
 } from "../utils/denomination";
@@ -31,7 +32,8 @@ const Create = () => {
     let receiveAmountRef: HTMLInputElement | undefined = undefined;
     let sendAmountRef: HTMLInputElement | undefined = undefined;
 
-    const { webln, denomination, wasmSupported, t } = useGlobalContext();
+    const { setDenomination, webln, denomination, wasmSupported, t } =
+        useGlobalContext();
     const {
         reverse,
         asset,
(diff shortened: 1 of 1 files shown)


kilrau (changes requested):
Second bullet point of https://github.com/BoltzExchange/boltz-web-app/issues/391 is missing: allow entering floats and autoswitch to btc if a `.` is entered. It needs both diretions, we had support cases the same way where people couldn't enter `0.1` as shown in sideswap.

kilrau:
> autoswitch to btc if a `.` is entered

Still not working (doesn't let me enter `.` when on sats denom). We need this direction too.

kilrau:
And conflicts need to be resolved

src/pages/Create.tsx
```diff
@@ -60,10 +62,21 @@ import { enableWebln } from "../utils/webln";
 const Create = () => {
     let receiveAmountRef: HTMLInputElement, sendAmountRef: HTMLInputElement;
 
+    // if btc and amount > 9, switch to sat
+    // user failed to notice the non satoshi denomination
+    const changeDenomination = (amount: number) => {
+        if (denomination() === denominations.btc && amount > 9) {
+            setDenomination(denominations.sat);
+        } else if (denomination() === denominations.sat && amount < 1) {
```
michael1011, line 70: This has the side effect of switching to BTC denom when the amount is cleared. Questionable if we want that

kilrau: We don't. No switching on clearing.

dni: is it better now?  :)

kilrau (changes requested):
Better but not there yet. when in sats deno and I enter `0.` it correctly switches to bitcoin and let's me enter. If I enter `1.` or `2.` etc it does weird things. 

[Screencast from 01-08-2024 07:28:14 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/05682578-3f9a-4f75-9e82-801ae6eefd42)

kilrau (changes requested):
My previous review comment is still not fixed. When I am in sat denomination and I enter `1.`  it jumps to `0.00000001`, should stay `1.`

michael1011 (approved):

PR #398, opened 2023-12-22
feat: slightly nicer denominaction btn

dni:
make a more obvious active state

screenshot-1703219176

closes: https://github.com/BoltzExchange/boltz-web-app/issues/390


dni:
closes #390

kilrau (changes requested):
This uses input field style (dotted line), doesnt change color on hover - just doesn't feel like a button yet.

Can we try a toggle style switch (which just toggles no matter where exactly you click). Sth like this but where the active side of the toggle has a brighter blue, the inactive a darker than the background:
image

kilrau (approved):
LGTM

PR #400, opened 2023-12-22
feat: swaplist alignment

dni:
align the swaplist in history and refundable swap


kilrau (changes requested):
This the alignment you wanted? Imo with fixed length for  ID and date better to ahve the right side buttons aligned too
image

dni:
good catch, i fucked up the css last second :) last year, not it is working again!

screenshot-1704279990

PR #408, opened 2024-01-03
chore: moving signals and ecpair to ts

dni:
skipLibCheck is needed for an error in `solid-start` which is used by the storage primitives

part 1. 
trying to make #341 easier to review.


package.json
```diff
@@ -13,7 +13,7 @@
         "testnet": "cp src/configs/config.testnet.ts src/config.ts; vite build; cp dist/index.html dist/404.html",
         "prettier": "prettier --write src tests",
         "prettier-check": "prettier --check src tests",
-        "tsc": "tsc",
+        "tsc": "tsc --skipLibCheck",
```
michael1011, line 16: Should be set in the tsconfig

michael1011 (changes requested):

PR #415, opened 2024-01-04
chore: move main components to ts

dni:
pr 4 of tsc update


src/Nav.tsx
```diff
@@ -11,7 +11,7 @@ import { setI18nConfigured } from "./signals";
 import "./style/nav.scss";
 
 const Nav = ({ network }) => {
-    let timeout;
+    let timeout: any;
```
michael1011, line 14:
```suggestion
    let timeout: number | undefined;
```

src/Pay.tsx
```diff
@@ -63,6 +63,7 @@ const Pay = () => {
         const tx = swapStatusTransaction();
 
         if (swap().asset === RBTC && tx && swap().claimTx === undefined) {
+            // @ts-ignore
```
michael1011, line 66: `swapStatusTransaction` has the wrong type

michael1011 (changes requested):

src/Pay.tsx
dni: fixed in follow up tsc pr's

PR #416, opened 2024-01-04
chore: move utils to ts

dni:
part 5


src/utils/feeChecker.ts
```diff
@@ -1,13 +1,13 @@
...
 
-const relevantMinerFees = (fees) => {
+const relevantMinerFees = (fees: any) => {
     return asset() === BTC
         ? fees.minerFees.quoteAsset
         : fees.minerFees.baseAsset;
 };
 
-export const feeChecker = (pairs) => {
+export const feeChecker = (pairs: any) => {
```
michael1011, line 10: Here we should use strong types. The format of the data is known

src/utils/denomination.ts
```diff
@@ -43,16 +50,20 @@ export const formatAmountDenomination = (denom, amount, fixed = false) => {
...
-    switch (denomination()) {
+export const convertAmount = (amount: number, denom: string): number => {
+    switch (denom) {
         case denominations.btc:
             const amountBig = new BigNumber(amount).multipliedBy(satFactor);
             return amountBig.toNumber();
         case denominations.sat:
             return Number(amount);
+        default:
+            return Number(amount);
```
michael1011, line 66:
```suggestion
        default:
            return Number(amount);
```

src/utils/denomination.ts
```diff
@@ -43,16 +50,20 @@ export const formatAmountDenomination = (denom, amount, fixed = false) => {
 
         case denominations.sat:
             return amount.toString();
+        default:
+            return amount.toString();
```
michael1011, line 54:
```suggestion
        default:
            return amount.toString();
```

tests/components/Fees.spec.tsx
```diff
@@ -8,6 +8,7 @@ import { cfg } from "../config";
 
 describe("Fees component", () => {
     beforeAll(() => {
+        // @ts-ignore
```
michael1011, line 11: The type of `setConfig` is wrong. Don't just ts-ignore here. That defeats the purpose of typescript

michael1011 (changes requested):

src/utils/feeChecker.ts
dni: open issue for that

PR #435, opened 2024-01-08
refactor: claim.ts and refund.ts

dni:
moved from helper.ts


tests/components/RefundButton.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+        expect(button.disabled).not.toBeTruthy();
+
+        // paste a invalid address
+        fireEvent.input(input, {
+            target: {
+                value: "XXXQ5FhU2497BryFfUgbqkAJE87aKHUhXMp",
+            },
+        });
+
+        expect(button.disabled).toBeTruthy();
```
michael1011, line 68: That should be a separate test case

src/components/RefundButton.tsx
```diff
@@ -0,0 +1,112 @@
...
+    setRefundAddress,
+    setRefundTx,
+} from "../signals";
+import { getAddress, getNetwork } from "../utils/compat";
+import { decodeInvoice } from "../utils/invoice";
+import { refund } from "../utils/refund";
+import { prefix0x, satoshiToWei } from "../utils/rootstock";
+import ContractTransaction from "./ContractTransaction";
+
+const RefundButton = ({ swap }: { swap: Accessor<Record<string, any>> }) => {
```
michael1011, line 21: The name is confusing
It's but just a button

src/components/SwapChecker.tsx
```diff
@@ -60,25 +61,30 @@ export const SwapChecker = () => {
...
+        const currentSwap = swaps().find((s) => activeSwap.id === s.id);
         if (
             currentSwap.claimTx === undefined &&
             data.transaction !== undefined &&
             (data.status === swapStatusPending.TransactionConfirmed ||
                 data.status === swapStatusPending.TransactionMempool)
         ) {
-            claim(currentSwap, data.transaction);
+            claim(currentSwap, data.transaction, (swap) => {
+                // TODO: this get the swap status updated with the claim tx
```
michael1011, line 84: ?

src/utils/claim.ts
```diff
@@ -0,0 +1,116 @@
...
+        true,
+        assetHash,
+        blindingKey,
+    );
+};
+
+export const claim = async (
+    swap: any,
+    swapStatusTransaction: { hex: string },
+    callback: (swap: any) => any,
```
michael1011, line 48: We should get rid of the callbacks. ***Especially*** in async functions

michael1011 (changes requested):

src/utils/claim.ts
dni: agreed, next pr

tests/components/RefundButton.spec.tsx
dni: done

src/components/SwapChecker.tsx
dni: done

tests/components/RefundButton.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+            },
+        });
+
+        const button = (await screen.findByTestId(
+            "refundButton",
+        )) as HTMLButtonElement;
+        expect(button.disabled).not.toBeTruthy();
+    });
+
+    test("button should be inactive after pasting in address", async () => {
```
michael1011, line 62:
```suggestion
    test("button should be inactive after pasting an invalid address", async () => {
```

michael1011 (approved):

PR #459, opened 2024-01-24
fix: undefined asset on refund page

michael1011:


kilrau (changes requested):
placeholder
image

kilrau (approved):
Looking good. Without currency in initial state and adds correct currency once I select a swap to refund. Can merge when tests passed.

dni (approved):

PR #464, opened 2024-02-05
feat: switch from SSE to WS

michael1011:
Depends on https://github.com/BoltzExchange/boltz-backend/pull/480


dni (changes requested):
test failing :D

dni (approved):
NICE

PR #467, opened 2024-02-07
fix: autoswitch off by 1

### LINKED ISSUE #452: Entering 9.X **while in BTC denomination** switches to sats keeping 9.X
Only could reproduce this with "9.X" while in BTC denomination
[Screencast from 01-22-2024 02:25:16 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/1c218cd9-23b5-41a2-8ee5-b58190eadd4c) so must be an edge case in https://github.com/BoltzExchange/boltz-web-app/pull/395

dni:
closes #452

we should actually only switch if its >= 10 so you cannot enter 9.x :)

### DIFF
--- a/src/pages/Create.tsx
+++ b/src/pages/Create.tsx
@@ -63,11 +63,11 @@ const Create = () => {
         minerFee,
     } = useCreateContext();
 
-    // if btc and amount > 9, switch to sat
+    // if btc and amount > 10, switch to sat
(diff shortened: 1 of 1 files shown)


kilrau (approved):
Still don't agree on switching to sats if I enter `.` but well yes this improves it

PR #471, opened 2024-02-10
feat: only show refund button when file is uploaded

### LINKED ISSUE #469: Refund: Hide address input box until swap selected/refund file uploaded
It looks like this takes input but it doesn't. Which is correct since we first need to know currency for validation.

image

dni:
closes #469

- dont use global signal for `refundTx`

### DIFF
--- a/.git-commit-template.txt
+++ b/.git-commit-template.txt
@@ -1,7 +1,7 @@
 # HEADER
 # |<----  Using a Maximum Of 50 Characters  ---->|
 <type>(<scope>): <subject>
-# Type can be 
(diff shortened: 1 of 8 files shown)


michael1011 (changes requested):
Missing tests

kilrau:
Testing if address field shows if a local storage swap is present brought two things to light:
- websocket on mainnet is not working yet
- not sure if this looks great (if there is a refundable swap in local storage), maybe add a bit of space and rename from `View` -> `Refund`
image

PR #488, opened 2024-02-22
feat: support boltz-client as backend

### LINKED ISSUE #492: Lazy loading
Web App is around 5.9MB, most of it dependencies. To reduce load time in remote places, we want to show a meaningful UI to select assets and enter swap amounts while libraries are still loading.

Needs a careful architecture to make sure we can't get into unexpected states when a button tries to use a loading library.

jackstar12:
closes #492


kilrau:
As discussed please move the lazy loading change into a separate PR

PR #489, opened 2024-02-24
feat: intermediate step when uploading refundjson and proper error

### LINKED ISSUE #470: Refund: Add intermediate redirect page if swap is still in progress
When uploading a refund file of a swap that isn't expired yet, we currently just jump to the swap progress page. The idea is good, but that's a big sus - we  have a "This swap is still in progress, redirecting to swap..." intermediate step to make the user aware why this is happening.

### LINKED ISSUE #481: Missing "Invalid Refund File" in swap button
Right now, user has no idea why address input and refund button are not working.

image

dni:
closes #470
closes #481

screenshot-1708772817

screenshot-1708774513

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -350,7 +350,7 @@ const dict = {
         testnet: "Testnet",
         broadcasting_claim: "Sende claim transaction...",
         open_swap: "Swap öffnen",
-        swap_in_progress: "Dieser Swap ist noch im Gange.",
+        swap_in_progress: "Dieser Swap ist noch nicht abgeschlossen.",
     },
     es: {
         language: "Español",
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -170,6 +170,8 @@ const dict = {
         branding: "Branding",
         testnet: "Testnet",
         broadcasting_claim: "Broadcasting claim transaction...",
+        open_swap: "Open Swap",
+        swap_in_progress: "This swap is still in progress.",
     },
     de: {
         language: "Deutsch",
@@ -347,6 +349,8 @@ const dict = {
         branding: "Branding",
         testnet: "Testnet",
         broadcasting_claim: "Sende claim transaction...",
+        open_swap: "Swap öffnen",
+        swap_in_progress: "Dieser Swap ist noch in Bearbeitung.",
     },
     es: {
         language: "Español",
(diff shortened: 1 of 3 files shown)


src/i18n/i18n.ts
```diff
@@ -170,6 +170,8 @@ const dict = {
         branding: "Branding",
         testnet: "Testnet",
         broadcasting_claim: "Broadcasting claim transaction...",
+        open_swap: "Open Swap",
+        swap_in_progress: "This swap is still in progress.",
```
michael1011, line 174: Super confusing. You can have swaps in progress even if they are not in local storage. This wording creates too many weird states for the user

src/pages/Refund.tsx
```diff
@@ -73,6 +80,7 @@ const Refund = () => {
                 })
                 .catch((e) => {
                     log.error("invalid file upload", e);
+                    setRefundInvalid(true);
                     input.setCustomValidity(t("invalid_refund_file"));
                 });
```
michael1011, line 85: Why don't we use `await` for both, the text file and the QR code scan, to avoid having the redundant logic?

src/pages/Refund.tsx
```diff
@@ -19,6 +19,8 @@ const Refund = () => {
     const { updateSwapStatus, wasmSupported, swaps, t } = useGlobalContext();
     const { setTimeoutEta, setTimeoutBlockheight } = usePayContext();
 
+    const [swapFound, setSwapFound] = createSignal(null);
```
michael1011, line 22: I don't like having this as signal. We will loose track of what signal does what in which combination

src/pages/Refund.tsx
```diff
@@ -143,6 +151,21 @@ const Refund = () => {
...
                     />
+                    <Show when={swapFound() !== null}>
+                        <hr />
+                        <p>{t("swap_in_progress")}</p>
+                        <button
+                            class="btn btn-success"
+                            onClick={() => navigate(`/swap/${swapFound()}`)}>
+                            {t("open_swap")}
+                        </button>
+                    </Show>
```
michael1011, line 162: Always the set `refundJson` and switch between open and refund if `swaps().some((swap) => swap.id === refundJson().id)`

src/pages/Refund.tsx
dni: we could but i think it fine, we may do something different for both in the future,, and the logging is different btw.

src/i18n/i18n.ts
kilrau: Looks like this was resolved. I find the current wording understandable.

michael1011 (approved):

kilrau (approved):

PR #496, opened 2024-03-09
feat: improve pasting

### LINKED ISSUE #305: Improve pasting
If satoshis is set and I am trying to paste a float (whole bitcoin), it doesn't work and it's not obvious why. Also if the amount is way out of max it doesn't work without any error message or anyting
- [x] I'd claim this happens every now and then, so we should detect if it's a valid float and auto switch to full bitcoin and paste
- [x] For any other illegal chars we should bring up our red "Could not paste, clipboard contains invalid characters" warning
- [x] throw error when trying to paste an >max digit limit amount

dni:
closes #305

it pretty though too differentiate between invalid amount  (max) or invalid characters on the regex, can we do 1 error message for all i think it is sufficient?

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -172,8 +172,8 @@ const dict = {
         broadcasting_claim: "Broadcasting claim transaction...",
         open_swap: "Open Swap",
         swap_in_progress: "This swap is still in progress.",
-        paste_invalid_chars:
-            "Could not paste, clipboard contains invalid characters",
+        paste_invalid:
+            "Clipboard contains invalid characters or maximum amount is exceeded",
     },
     de: {
         language: "Deutsch",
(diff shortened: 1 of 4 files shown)


kilrau (changes requested):
- [x] the generic message is good enough :+1: 
image

- [ ] pasting a float that is within min/max works
image
but instead of throwing it somehow breaks the autoswitch to btc denomination if paste amount is full bitcoin and >max *while in sats denomination*
image

dni:
> Still not fixed for me: image

sorry, i accidently requested a review to early :)

kilrau (changes requested):
Previous issue fixed. One more tiny thing: if I try pasting with cursor in the middle of this amount where max digits of the input box are reached it should throw the error as well. Right now it just doesn't paste.

image

kilrau (approved):
LGTM

PR #497, opened 2024-03-10
refactor: introduce types.ts and move signals to contexts

dni:
- introduce a types.ts for all the types
- move buttonLabel signal to context


src/types.ts
```diff
@@ -0,0 +1,4 @@
+export type buttonLabelParams = {
```
michael1011, line 1: Types should be PascalCase

michael1011 (approved):

Issue #443, opened 2024-01-10
Enforce BOTH amount boxes were calculated & populated before resolving any LNURL/LNADDRESS


kilrau:
Enforce BOTH amount boxes were populated with calucalated amounts before resolving any LNURL/LNADDRESS. Right now web app can get into states where it resolves lnurl/lnaddresses to invoices if only one amount is set. That's dangerous since it's not clear what amount invoice has. 

Requirement that this is rock solid before we do https://github.com/BoltzExchange/boltz-web-app/issues/441

dni:
this is solved by https://github.com/BoltzExchange/boltz-web-app/pull/500

Issue #353, opened 2023-12-08
Normal Swaps: Change input validation order to avoid default error state


kilrau:
Tested latest master:

When I open boltz.exchange and my last swap direction was a reverse swap, I have a pretty ugly error state with invoice input field red circled and button complaining about invalid invoice
![image](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/66f73d60-e45f-4b4c-99a0-0422484b3f83)

Two things to be done:
- [x] Rewording to `Please enter a valid lightning invoice...`
- [x] Rewording to `Please enter a valid {asset} address`
- [x] Don't circle invoice input field red if amount is 0 (which is default). Meaning: order of input validation should always be amount first then invoice/btc address. Reverse Swaps already behave how it should be, let's do the same for normal swaps!

[Screencast from 12-22-2023 02:31:27 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/8820409f-b185-4c0d-a5f5-bf340a58a4cc)

PR #500, opened 2024-03-11
fix: bail validation and refactor create button labels

dni:
- button label is now way easier to work with
- validation only happens if there is a amount set at all

### DIFF
--- a/src/components/CreateButton.tsx
+++ b/src/components/CreateButton.tsx
@@ -88,6 +88,7 @@ export const CreateButton = () => {
                 lnurl,
                 online,
                 minimum,
+                asset,
             ],
             () => {
                 if (!online()) {
--- a/src/components/CreateButton.tsx
+++ b/src/components/CreateButton.tsx
@@ -1,9 +1,10 @@
 import { useNavigate } from "@solidjs/router";
+import BigNumber from "bignumber.js";
 import { crypto } from "bitcoinjs-lib";
 import { OutputType } from "boltz-core";
 import { randomBytes } from "crypto";
 import log from "loglevel";
-import { createEffect, createMemo, createSignal } from "solid-js";
+import { createEffect, createMemo, createSignal, on } from "solid-js";
 
 import { BTC, RBTC } from "../consts";
 import { useCreateContext } from "../context/Create";
@@ -11,6 +12,7 @@ import { useGlobalContext } from "../context/Global";
 import { useWeb3Signer } from "../context/Web3";
 import { ButtonLabelParams } from "../types";
 import { getPairs } from "../utils/boltzClient";
+import { formatAmount } from "../utils/denomination";
 import { ECPair } from "../utils/ecpair";
 import { fetcher, getPair } from "../utils/helper";
 import { extractAddress, fetchLnurl } from "../utils/invoice";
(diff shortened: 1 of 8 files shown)


src/components/CreateButton.tsx
```diff
@@ -34,42 +45,108 @@ export const CreateButton = () => {
...
+                        params: {
+                            amount: formatAmount(
+                                BigNumber(lessThanMin ? minimum() : maximum()),
+                                denomination(),
+                            ),
+                            denomination: denomination(),
+                        },
+                    });
+                    return;
+                } else {
```
michael1011, line 114: No need for else if you always return in the previous `if`s

src/components/CreateButton.tsx
```diff
@@ -209,7 +286,7 @@ export const CreateButton = () => {
             id="create-swap-button"
             data-testid="create-swap-button"
             class={buttonClass()}
-            disabled={buttonDisable() || !online()}
+            disabled={buttonDisable()}
```
michael1011, line 289: image
Invoices of the wrong network have a yellow button that can't be clicked. How about setting the class of the button when the disabled status is changed?

michael1011 (changes requested):
There are no tests

dni:
closes #443

dni:
closes #353

tests/components/CreateButton.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+        signals.setAmountValid(true);
+        signals.setAddressValid(true);
+        signals.setReverse(true);
+        const btn = (await screen.findByText(
+            i18n.en.create_swap,
+        )) as HTMLButtonElement;
+        expect(btn).not.toBeUndefined();
+        expect(btn.disabled).toBeFalsy();
+        signals.setAddressValid(false);
+        expect(btn.disabled).toBeTruthy();
```
michael1011, line 94: Let's also assert the text changed to address invalid

michael1011 (approved):

kilrau (changes requested):
https://github.com/BoltzExchange/boltz-web-app/issues/443 seems to be fixed, couldn't get web app to resolve my lightning address with invalid/empty amounts

https://github.com/BoltzExchange/boltz-web-app/issues/353 for empty amounts the "Please provide a valid LN invoice, LNAddress or LNURL" should be shown too if I enter sth invalid: 
image
Same for Normal Swaps with empty amounts, it should show "Invalid BTC address"
image

kilrau (changes requested):
Let's fix error prios:
- let's stick with "Minimum amount is" as default state when loading the page, just as current master. But NEVER circle invoice/address field red if it's empty. Only when it's an invalid lnurl/lnaddress format. Grey button with error message is clear enough. Yes WebLN can be used to paste an invoice with amount and users will click the button if it's there even though the message says enter an amount. Accepted edge case trade-off.
image
- amount error supersedes all other errors, wrong ln address/lnurl show only if valid amount is set, test: amount error stays when I enter wrong lnurl/address
- invoice errors like 0 amount can show as-is as they are popups
- re-run amount validation when I manipulate amounts that passed validation before
- re-run lnurl/lnaddress/invoice/address validationwhen I manipulate an input that passed validation before

kilrau (changes requested):
One minor things left:

- [ ] you are still showing red if the address input field is empty. Please show grey in case it's empty. Shouldn't have an error state if I'm just done punching in amounts
image

kilrau (approved):
LGTM

michael1011 (approved):

PR #503, opened 2024-03-13
feat: switch assets based on address

### LINKED ISSUE #374: UX: Switch assets based on address/invoice input
When a Liquid address is entered/scanned, the frontend should switch the receiving symbol the L-BTC. Similar concept for Bitcoin addresses and lightning invoices/lnurls/sat domains

dni:
closes #374

switches onchain asset based on address paste


src/utils/compat.ts
```diff
@@ -83,6 +86,33 @@ const decodeAddress = (asset: string, addr: string): DecodedAddress => {
...
+    });
+    liquidPrefixes.forEach((prefix: string) => {
+        if (address.startsWith(prefix)) {
+            asset = LBTC;
+        }
+    });
+    if (asset !== null) {
+        return asset;
+    }
+    // fallback to more costly decoding
```
michael1011, line 104: Did you measure the performance of decoding the address?

src/utils/compat.ts
```diff
@@ -26,7 +26,10 @@ import {
 import { Network as LiquidNetwork } from "liquidjs-lib/src/networks";
 
 import { network } from "../config";
-import { LBTC } from "../consts";
+import { BTC, LBTC } from "../consts";
+
+const btcPrefixes = ["bc", "tb", "bcrt"];
+const liquidPrefixes = ["lq", "ex", "tlq", "tex", "el", "ert"];
```
michael1011, line 32: Don't hardcode those; if anything, get them from the configure network. Also those are only the b(l)ech32(m) ones?

src/utils/compat.ts
```diff
@@ -83,6 +86,33 @@ const decodeAddress = (asset: string, addr: string): DecodedAddress => {
...
+    }
+    // fallback to more costly decoding
+    try {
+        decodeAddress(BTC, address);
+        return BTC;
+    } catch (e) {}
+    try {
+        decodeAddress(LBTC, address);
+        return LBTC;
+    } catch (e) {}
```
michael1011, line 112: We need to decode the address anyways. Why don't we start decoding with the set network and then try all the others known to us?

- if address is valid for set network -> address valid
- address is valid for other network -> address valid & switch network
- address is not valid for any network -> address invalid

src/components/AddressInput.tsx
```diff
@@ -1,32 +1,54 @@
...
         const inputValue = input.value.trim();
         const address = extractAddress(inputValue);
+        const invoice = extractInvoice(inputValue);
+
+        // auto switch direction based on invoice
+        if (isInvoice(invoice)) {
+            setAssetReceive(LN);
+            setAssetSend(asset());
+            setInvoice(invoice);
+            notify("success", t("switch_paste"));
```
michael1011, line 37: Not a fan of throwing a notification for every action you do

michael1011 (changes requested):

src/components/AddressInput.tsx
dni: i like it, and i talked with @kilrau about it.

src/utils/compat.ts
dni: i did not measure it, but noticed the browser hanging sometimes decoding the address

src/utils/compat.ts
dni: we  dont need to decode the address when pasting into the `InvoiceInput` and `setOnchainAddress` would decode it twice

kilrau:
I think this branch needs a rebase, I have rootstock dangling in there 
image

src/i18n/i18n.ts
```diff
@@ -707,6 +712,7 @@ const dict = {
         swap_in_progress: "此交换仍在进行中。",
         paste_invalid: "剪贴板包含无效字符或超出最大金额",
         email: "邮箱",
+        switch_paste: "根据粘贴的内容切换交换方向/资产",
```
michael1011, line 715: @kilrau check the translations

michael1011 (approved):

src/i18n/i18n.ts
kilrau: LGTM

kilrau (changes requested):
Notification of switch doesn't always show, e.g. here when switching back from lightning to liquid:
[Screencast from 03-20-2024 11:39:06 AM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/5ffa23a6-cd60-41a5-bbc2-c2a8b274ebdc)

kilrau (approved):
LGTM

PR #510, opened 2024-03-19
feat: use `solid-icons` instead of single svgs

dni:
saves time when adding new icons.

search icons here: https://solid-icons.vercel.app/


michael1011 (approved):

kilrau (approved):
looking good

PR #523, opened 2024-03-20
feat: invalid refund address when using lock address

dni:


src/components/RefundButton.tsx
```diff
@@ -100,6 +100,11 @@ const RefundButton = ({
     const refundAddressChange = (evt: InputEvent, asset: string) => {
         const input = evt.currentTarget as HTMLInputElement;
         const inputValue = input.value.trim();
+        if (inputValue === swap().address) {
+            log.debug("don't refund to lock address");
```
michael1011, line 104:
```suggestion
            log.debug("refunds to lockup address are blocked");
```

michael1011 (approved):

Issue #346, opened 2023-12-05
0-conf toggle & notifications


kilrau:
We moved web app to full [0-conf back in June](https://github.com/BoltzExchange/boltz-web-app/issues/92 ) to make sure reverse swaps, especially on mobile, don't fail. Users rightfully recently criticized us for that, we should offer users the option for web app to wait for one confirmation for the lockup tx with a toggle just like we had before. We also want to ask user for notification rights when toggled on and send a notification once lockup tx confirmed to open browser tab and make sure claim succeeds.

Special: on iOS browser push notifications only work for PWAs on home screen:
https://developer.apple.com/documentation/usernotifications/sending_web_push_notifications_in_web_apps_and_browsers

kilrau:
[Here](https://www.youtube.com/watch?v=N9zpRvFRmj8) a nice video of Google explaining how sending and receiving push messages work. Important takeaway: the subscription id part of the unique endpoint url enables linking swaps that used it. Not great for user's privacy.

And obviously it won't work [on phones without Google Play Services installed](https://discuss.grapheneos.org/d/3141-how-do-push-notifications-from-websites-work/4) and we expect to have issues with privacy browsers too as they probably stubbed the push service.

Tl;dr: not looking like a viable solution to our problem of waking up user's browsers to make claiming succeed

kilrau:
We should settle how to handle this @dni @michael1011

PR #528, opened 2024-03-22
feat: add browser notification

dni:
used when a swap is claimed or refunded.

screenshot-1711525563

screenshot-1714982106

now it  is opt in only via settings


kilrau (changes requested):
I open the landing page and it immediately asks for notification permissions - NO
image

Then again on "Create Swap"... 

Please read the issue again! We only want notification permissions *if user toggles 0-conf off*: https://github.com/BoltzExchange/boltz-web-app/issues/346 so that should be the trigger.

kilrau:
As our testing just revealed: browser notification is "in your face" on page load, only should be triggered on toggle

kilrau:
In your face permission request gone, please resolve conflicts and request review, I'll do some e2e testing @dni

kilrau (changes requested):
- need to gracefully handle incognito window
[Screencast from 2024-05-22 13-26-16.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/2ca3be57-952a-4a52-b67a-6841d7055a75)
- didn't get any notification when I was in a different tab while the swap completed even though notifications were turned on

Tested in FF & Chromium.

kilrau:
Edge case but if the fix is easy let's please do it: if I withdraw notification permissions while switch is on, it's still on after reload. Reload should check permissions I guess.

[Screencast from 2024-05-22 15-43-46.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/5b65001e-0af9-4f20-b01f-a4e83595a719)

kilrau (changes requested):
We fixed a couple of things in my regtest setup and notfications are working now.

TODOs:

1. what I mentioned [above](https://github.com/BoltzExchange/boltz-web-app/pull/528#issuecomment-2124840924):
- if user clicks "block" on initial permission request
image
- OR user is in incognito window
[Screencast from 2024-05-22 16-33-12.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/ba3ec599-7aae-4127-9c90-4047668a67d8)
Both resulting in permissions not granted, let's throw "Notification permissions not granted." error when user tries to enable here. Right now just nothing happens when trying to enable notifications in these states.

2. Notifications for all failed swaps, right now it seems there are no notifications for failed swap states:
Screenshot from 2024-05-22 16-37-05

kilrau (approved):
Previous review was addressed, tested in FF&Chromium and works as expected. Strings fixed.

michael1011 (changes requested):

kilrau:
> 2\. Notifications for all failed swaps, right now it seems there are no notifications for failed swap states:

Did you add notifications for failed swaps? I broke my regtest again and can't test rn.

kilrau:
> 2\. Notifications for all failed swaps, right now it seems there are no notifications for failed swap states:

Moved into separate issue https://github.com/BoltzExchange/boltz-web-app/issues/601, from my side ok to merge this without

src/i18n/i18n.ts
```diff
@@ -182,15 +183,19 @@ const dict = {
...
         logs: "Logs",
-        logs_tooltip: "Logs of the web app, useful for debugging.",
-        enable_audio_notifications: "Enable Audio Notifications",
+        logs_tooltip: "Logs of the web app, useful for debugging",
+        enable_audio_notifications: "Audio Notifications",
         enable_audio_notifications_tooltip:
             "Enable or disable audio notifications",
         on: "on",
         off: "off",
+        notification_header: "Boltz Swap",
```
michael1011, line 195:
```suggestion
        notification_header: "Boltz",
```
I'd call this just Boltz. It's cleaner

michael1011 (approved):

src/i18n/i18n.ts
kilrau: Cleaned. For all languages.

PR #532, opened 2024-03-23
chore: use compareTrees function from boltz-core

michael1011:
This PR updates some dependencies, including boltz-core to be able to use the new [`compareTrees`](https://github.com/BoltzExchange/boltz-core/commit/ce129faed3c3914358e5b697a4d1e4d612e711e7) function in v2.1.2


dni (approved):
LGTM

PR #534, opened 2024-03-27
bug: catch exception on refund page

dni:
screenshot-1711526219

happened if status returned a 404, request errors are already logged so do nothing. ps. this should never happen in production :)

### DIFF
--- a/src/pages/Refund.tsx
+++ b/src/pages/Refund.tsx
@@ -131,8 +131,8 @@ const Refund = () => {
                         await getSubmarineTransaction(swap.asset, swap.id);
                         addToRefundableSwaps(swap);
                     }
-                } catch (_) {
-                    log.warn("failed to get swap status", swap.id);
+                } catch (e) {
+                    log.warn("failed to get swap status", swap.id, e);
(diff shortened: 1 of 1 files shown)


src/pages/Refund.tsx
```diff
@@ -116,19 +116,23 @@ const Refund = () => {
...
+                        await getSubmarineTransaction(swap.asset, swap.id);
                         addToRefundableSwaps(swap);
-                        return;
                     }
-
-                    // Make sure coins were locked for the swap with status "swap.expired"
-                    await getSubmarineTransaction(swap.asset, swap.id);
-                    addToRefundableSwaps(swap);
+                } catch (_) {
+                    // do nothing, there is already a request error in the console
```
michael1011, line 135: An error of a failed HTTP request, if that happens to be enabled in your browser
Add a warning log

michael1011 (changes requested):

src/pages/Refund.tsx
dni: done

PR #542, opened 2024-03-28
fix: claim duplications

### LINKED ISSUE #509: Rebroadcasting claims causes errors
Looks like this is a bit too aggressive
image
image

dni:
closes #509 

the effect in TxClaimed was triggering the claim, i think this is not necessary because the sweapchecker handles it. tested it manually and it works.

@michael1011 i think you introduced that any reasoning behind that?

introduction of claim queue so websocket updates do not overlap each other


michael1011 (approved):

tests/utils/lock.spec.ts
```diff
@@ -0,0 +1,31 @@
+import Lock from "../../src/utils/lock";
+
+describe("Lock", () => {
+    test("should not run promises concurrently", async () => {
```
dni, line 4: sequentially

PR #543, opened 2024-03-29
bug: remove spaces from copy sat amount

dni:
should have spaces in sat amount when copying.
spaces can be safely replaced for all copies, its never valid

### DIFF
--- a/src/components/CopyButton.tsx
+++ b/src/components/CopyButton.tsx
@@ -10,7 +10,7 @@ const CopyButton = ({ data, label }) => {
     const [buttonClass, setButtonClass] = createSignal<string>("btn");
     const [buttonActive, setButtonActive] = createSignal<boolean>(false);
     const onClick = () => {
-        clipboard(data);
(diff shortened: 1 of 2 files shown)


src/components/CopyButton.tsx
```diff
@@ -10,7 +10,7 @@ const CopyButton = ({ data, label }) => {
     const [buttonClass, setButtonClass] = createSignal<string>("btn");
     const [buttonActive, setButtonActive] = createSignal<boolean>(false);
     const onClick = () => {
-        clipboard(data);
+        clipboard(data.replaceAll(" ", ""));
```
michael1011, line 13: What if we ever want to copy something that needs to contain a space? Could we only apply this to numbers?

michael1011 (changes requested):
No test

src/components/CopyButton.tsx
dni: right now spaces are never valid and `50 000` is also not a number

michael1011 (approved):

PR #548, opened 2024-04-02
feat: add usecase section to homepage

### LINKED ISSUE #297: Add "Use Case" section to landing page
We'd want to add a section about "Use Cases" - what one can do with Boltz - with the following content to the landing page.

🫰 Reliably cheap LN channel rebalancing
❄️ DCAing from LN into cold storage
👨‍💻 Increase coin privacy from CEX purchases

Probably adding it as one of these boxes is a good idea. But need to try what looks good.
image

dni:
a different approach to the previous try.

closes #297

screenshot-1712037277
screenshot-1712037271

### DIFF
--- a/src/pages/Hero.tsx
+++ b/src/pages/Hero.tsx
@@ -128,6 +128,22 @@ export const Hero = () => {
                     </div>
                 </div>
 
+                <h2 class="headline">Use cases</h2>
+                <div id="cases">
+                    <div class="case">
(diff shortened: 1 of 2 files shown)


kilrau:
If you agree, let's just not add the use case section and close issue + PR. We have way too many use cases by now and a two liner + icon could never really explain the advanced things you can do with Boltz anyways.

PR #549, opened 2024-04-03
feat: add setting menu

dni:
adding settings menu, with denomination and decimal separator, in the future will have more like, 0conf, feerate, etc.

screenshot-1714730885
screenshot-1714730882

### DIFF
--- a/src/components/SettingsMenu.tsx
+++ b/src/components/SettingsMenu.tsx
@@ -19,7 +19,7 @@ const SettingsMenu = () => {
             <span class="close" onClick={() => setSettingsMenu(false)}>
                 <IoClose />
             </span>
-            <hr />
+            <hr class="spacer" />
             <span class="setting">
                 <label>{t("denomination")}: </label>
                 <Tooltip label="denomination_tooltip" />
--- a/src/components/SettingsMenu.tsx
+++ b/src/components/SettingsMenu.tsx
@@ -1,0 +1,39 @@
+import { IoClose } from "solid-icons/io";
+
+import { useGlobalContext } from "../context/Global";
+import "../style/settings.scss";
+import Denominaton from "./Denomination";
+import Separator from "./Separator";
+import Tooltip from "./Tooltip";
+
+const SettingsMenu = () => {
+    const { t, settingsMenu, setSettingsMenu } = useGlobalContext();
+
+    return (
+        <div
+            id="settings-menu"
+            class="frame assets-select"
+            onClick={() => setSettingsMenu(false)}
+            style={settingsMenu() ? "display: block;" : "display: none;"}>
+            <h2>{t("settings")}</h2>
+            <span class="close" onClick={() => setSettingsMenu(false)}>
+                <IoClose />
+            </span>
+            <hr />
+            <span class="setting">
+                <label>Denominaton: </label>
+                <Tooltip label="denomination_tooltip" />
+                <div class="spacer"></div>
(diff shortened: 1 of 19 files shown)


kilrau:
Looking good - I know still a mock but should remove denomination switch when it's in settings

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,49 @@
...
+    let timeout: any = null;
+
+    const tooltipClick = (evt: MouseEvent) => {
+        const target = evt.currentTarget as HTMLSpanElement;
+        (target.parentNode as HTMLSpanElement).classList.add("active");
+        timeout = setTimeout(() => {
+            (target.parentNode as HTMLSpanElement).classList.remove("active");
+        }, TIMEOUT_CLICK);
+        evt.stopPropagation();
+    };
```
jackstar12, line 21: feels a bit unnecessary

src/pages/History.tsx
```diff
@@ -77,6 +79,7 @@ const History = () => {
     return (
         <div id="history">
             <div class="frame">
+                <SettingsCog />
```
jackstar12, line 82: i wouldnt put it on refund / history as long as there's no settings affecting those pages

src/components/SettingsMenu.tsx
```diff
@@ -0,0 +1,39 @@
...
+import Tooltip from "./Tooltip";
+
+const SettingsMenu = () => {
+    const { t, settingsMenu, setSettingsMenu } = useGlobalContext();
+
+    return (
+        <div
+            id="settings-menu"
+            class="frame assets-select"
+            onClick={() => setSettingsMenu(false)}
```
jackstar12, line 16: Not a fan of this. I would rather expect it to close on an outside click

src/components/Tooltip.tsx
dni: its for mobile devices, hover wont work well there

src/pages/History.tsx
dni: there may be in the future, so i put it on all pages where the box is.

src/components/SettingsMenu.tsx
dni: basically the same from assetselect, the pop over has an :after element fullsize which darkens everython in the background, if click anywaere else but the relevant inputs the popover will close

src/pages/History.tsx
jackstar12: of course, but there arent any yet. so id say only add id there when theres relevant settings

jackstar12 (approved):

src/components/SettingsMenu.tsx
jackstar12: > if clicked anywhere else but the relevant inputs the popover will close

which feels weird to me, but I wont die on this hill

kilrau: I agree it's weird and it's only "working" on asset select because the actionable buttons are that large that it's hard to misclick. I agree it should only close on [x] and clicking outside, let's change behavior for both.

kilrau: > I agree it's weird and it's only "working" on asset select because the actionable buttons are that large that it's hard to misclick. I agree it should only close on [x] and clicking outside, let's change behavior for both.

Can be done in a separate PR if you prefer

kilrau (approved):

PR #551, opened 2024-04-09
feat: chain swaps

michael1011:
- adds chainswaps (L-BTC to BTC and BTC to L-BTC) 
- big refactor to make not only submarine swaps possible


kilrau (changes requested):
Hmm that's a valid liquid testnet address from green

image
image

Bitcoin Testnet refund works as expected. So something is off with Liquid testnet address validation.

kilrau:
Also leaving this one here, even though I am still struggling to find a way to reproduce. Not a biggie in general as I can simply select one of the two and switch to another valid asset but I'd love to know how to get there.

[signal-2024-04-22-164336_002.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/85372ee0-3558-4623-8683-7aa6e7a6510d)

kilrau:
The swap in this refund file is the same as the one that shows in local storage on top and is refundable now, so it should go to the refund page and not display the "swap still in progress" page. Probably not related to this PR, just lmk @michael1011  and I'll put it into a separate issue for @dni 
Screenshot from 2024-04-23 16-05-30

kilrau:
And as @dni just noticed:

made a chainswap, downloaded json,  paid to little > fail > deleted localstorage > now tells me invalid refund file

kilrau:
Gave this another run on regtest. Observations/questions:
- How do I configure a separate fee for chain swaps via `boltz.conf`? Currently it seems to be taking 0.25% from Liquid Reverse Swaps for both sides. I know this is because of our legacy and it doesn't have to be nice, but we need to be able to set separate fees for all chain swaps *per direction* too.
- I adjusted `boltz.conf` to match mainnet values, but whatever I do min amount for chain swaps seems to be hardcoded to `102000` sats - what's that about?
- Network fee stays the same when I switch directions, this definitely shouldn't be
[Screencast from 2024-05-13 15-01-17.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/7b8fd160-5c7d-4b13-871b-21cd9fd89572)

kilrau (changes requested):
Can't refund on Liquid regtest because of "invalid value prefix" error
Screenshot from 2024-05-13 15-59-15

kilrau (changes requested):
Liquid refunds work as expected now! :tada: 

- [ ] Probably not related to this PR, but when I try to refund a swap after restarting regtest, backend obviously throws `warn: Request GET /v2/swap/chain/35qC5AcUFENP/transactions failed: {"error":"No such mempool or blockchain transaction. Use gettransaction for wallet transactions. ID: 01b8dec4361179b221519c9f28938e68ae90050c5fe2dafb852b35ad331c4320"}
` web app doesn't show anything and refund button just hangs there. Web App should show a `No such mempool or blockchain transaction` error.

dni (approved):
LGTM

dni:
closes #585

kilrau:
Arrow looks alright

kilrau (approved):
- image

Invalid pair error works as well as expected, superseding all other errors

- json refunds in an incognito tab without any local storage info works as expected too

dni (approved):

src/context/Global.tsx
```diff
@@ -115,6 +117,29 @@ const migrateSwapsFromLocalStorage = async () => {
...
+        if (!("type" in swap)) {
+            if (swap.reverse) {
+                swap.type = SwapType.Reverse;
+                swap.assetSend = LN;
+                swap.assetReceive = swap.asset;
+            } else {
+                swap.type = SwapType.Submarine;
+                swap.assetSend = swap.asset;
+                swap.assetReceive = LN;
+            }
```
michael1011, line 134: Did you ever test this? In progress swaps need to go through and there are bunch of other properties we renamed (eg keys)

src/context/Global.tsx
```diff
@@ -115,6 +117,29 @@ const migrateSwapsFromLocalStorage = async () => {
     return migratedSwapCount;
 };
 
+const migrateToChainSwaps = async (swapsForage: any) => {
+    const swaps = await swapsForage.keys();
+    let migratedSwapCount = 0;
```
michael1011, line 122: Use `swaps.length`

src/context/Global.tsx
```diff
@@ -201,12 +229,49 @@ const GlobalProvider = (props: { children: any }) => {
...
+                    case "reverse":
+                        lookup[LN] = lookup[LN].concat(pairData);
+                        break;
+                    case "submarine":
+                        lookup[pair].push(LN);
+                        break;
+                }
+            }
+        }
+        return lookup;
```
michael1011, line 265: Why not just use Sets?

src/context/Global.tsx
```diff
@@ -260,14 +325,25 @@ const GlobalProvider = (props: { children: any }) => {
             );
         });
 
-    const setSwapStorage = (swap: SwapWithId) =>
+    migrateToChainSwaps(swapsForage)
```
michael1011, line 328: This would run and iterate over all swaps on every reload. How about introducing versioning?

src/i18n/i18n.ts
```diff
@@ -608,6 +610,7 @@ const dict = {
         browsernotification_tooltip:
             "Activar o desactivar notificaciones del navegador",
         browsernotification_error: "Permisos de notificación denegados",
+        invalid_pair: "Par no válido",
```
michael1011, line 613: @kilrau did you check these?

src/pages/Pay.tsx
```diff
@@ -22,22 +31,105 @@ import TransactionConfirmed from "../status/TransactionConfirmed";
...
+    swapStatus,
+    contractTransaction,
+    contractTransactionType,
+}: {
+    swap: Accessor<SomeSwap>;
+    swapStatus: Accessor<string>;
+    contractTransaction: Accessor<string>;
+    contractTransactionType: Accessor<TransactionType>;
+}) => {
+    // Refund transactions are handled SwapRefunded
```
michael1011, line 58: What is that supposed to mean?

src/utils/boltzClient.ts
```diff
@@ -3,18 +3,22 @@ import { Musig } from "boltz-core";
 import { Buffer } from "buffer";
 import { Transaction as LiquidTransaction } from "liquidjs-lib";
 
+import { config } from "../config";
+import { SwapType } from "../consts/Enums";
 import { fetcher } from "./helper";
 
+const cooperativeErrorMessage = "Cooperative swaps are disabled";
```
michael1011, line 10: What are "cooperative swaps"?
```suggestion
const cooperativeErrorMessage = "cooperative signatures for swaps are disabled";
```

src/utils/claim.ts
```diff
@@ -131,102 +135,226 @@ const claimTaproot = async (
...
-        return;
-    }
+const claimChainSwap = async (
+    swap: ChainSwap,
+    lockupTx: TransactionInterface,
+    cooperative = true,
+): Promise<TransactionInterface> => {
+    log.info(`Claiming Chain swap cooperatively: ${cooperative}`);
+    //
+    // Create our claim transaction
```
michael1011, line 149:
```suggestion
```

src/components/AssetSelect.tsx
```diff
@@ -36,34 +33,34 @@ const SelectAsset = () => {
...
     const isSelected = (asset: string) => {
         return (
             asset ===
-            (assetSelected() === sideSend ? assetSend() : assetReceive())
+            (assetSelected() === Side.Send ? assetSend() : assetReceive())
         );
     };
 
+    createEffect(() => {
+        if (!("LN" in pairsLookup())) return;
```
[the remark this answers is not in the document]
dni: pairlookup only will be populated when fetchPairs is called so the keys like `LN` would be undefined on first pagelaod

src/components/InvoiceInput.tsx
```diff
@@ -89,7 +91,7 @@ const InvoiceInput = () => {
 
     createEffect(
         on([amountValid, invoice], () => {
-            if (!reverse() && asset() !== RBTC) {
+            if (swapType() === SwapType.Submarine && assetReceive() !== RBTC) {
```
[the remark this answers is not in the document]
dni: its `assetReceive() !== RBTC` so everything but RBTC?

src/i18n/i18n.ts
kilrau: Yes. ES was ok, for CN I did a small fix: https://github.com/BoltzExchange/boltz-web-app/pull/551/commits/dcfce3a8bbc6ddb28e6c20588aa08b7b43338d59

PR #559, opened 2024-04-22
feat: show qr code icon depending on invoice/onchain asset

### LINKED ISSUE #558: UI feedback: Boltz icon in QR code confusing when QR contains a Liquid address
This is a minor UI feedback.

When doing an `LBTC -> lightning` swap, the Boltz icon at the center of the QR code led me to think the QR was an invoice, while it is actually an LBTC address.

Screenshot 2024-04-22 alle 15 25 05

dni:
closes #558

screenshot-1713796813
screenshot-1713796809
screenshot-1713796805

### DIFF
--- a/src/components/QrCode.tsx
+++ b/src/components/QrCode.tsx
@@ -6,6 +6,7 @@ import "../style/qrcode.scss";
 
 interface QrCodeProps {
     data: string;
+    asset?: string;
 }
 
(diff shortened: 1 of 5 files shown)


kilrau (approved):
Works, LGTM

kilrau (changes requested):
Hmm Rootstock doesn't show any QR doesn't it - so not sure why we have qr logos for it.

image

PR #576, opened 2024-05-05
feat: add error reporting

dni:
- notify user when claiming failed
- i18n error messages
- add errors table to indexed db
- add ui into settings with download / copy errors

screenshot-1714894236


kilrau:
Very nice!

- Maybe let's go for "copy" only for now, no download option, as download won't work on iOS without shenanigans (where we crucially need it)? Or have download on Desktop only.
- I'd rename to `Local Errors:` to make it crystal clear that nothing is being sent anywhere
- show all functions as icons as space is very limited and it feels cluttered with things written out.
- Padding is off on mobile
image

kilrau (changes requested):
Very nice!

- Maybe let's go for "copy" only for now, no download option, as download won't work on iOS without shenanigans (where we crucially need it)? Or have download on Desktop only.
- I'd rename to `Local Errors:` to make it crystal clear that nothing is being sent anywhere
- show all functions as icons as space is very limited and it feels cluttered with things written out.
- Padding is off on mobile
image

src/i18n/i18n.ts
```diff
@@ -178,9 +178,18 @@ const dict = {
         canary: "Canary",
         settings: "Settings",
         decimal_separator: "Decimal Separator",
-        denomination_tooltip: "Choose your preferred denomination: BTC or sats",
-        decimal_tooltip:
-            "Choose your preferred decimal separator: dot or comma",
+        denomination_tooltip: "Choose your preferred denomination: btc or sat",
```
michael1011, line 181: Why lowercase BTC?

src/components/ErrorLog.tsx
```diff
@@ -0,0 +1,59 @@
...
+        const reports = await getReports();
+        clipboard(JSON.stringify(reports));
+    };
+
+    const download = async (evt: MouseEvent) => {
+        evt.stopPropagation();
+        downloadJson(
+            `boltz-errors-${Math.floor(Date.now() / 1000)}`,
+            await getReports(),
+        );
```
michael1011, line 37: That is not gonna work on Safari on mobile

src/components/ErrorLog.tsx
```diff
@@ -0,0 +1,59 @@
...
+            &nbsp;
+            <span onClick={download} class="btn-small">
+                {t("download")}
+            </span>
+            &nbsp;
+            <span onClick={copy} class="btn-small">
+                {t("copy")}
+            </span>
+            &nbsp;
+            <span class="btn-small btn-danger">{count()}</span>
```
michael1011, line 54: Why show the count explicitly?

src/components/RefundButton.tsx
```diff
@@ -169,15 +168,18 @@ const RefundButton = ({
                             msg = t("locktime_not_satisfied");
                         }
                         log.error(msg);
-                        setNotification(msg);
+                        notify("error", msg);
+                        report(swap().id, msg, jsonError);
```
michael1011, line 172: I am afraid that catching errors that are being thrown here is not going to be good enough. Can we hook into the logging library and also record every log statement?

src/i18n/i18n.ts
kilrau: I already fixed this here: https://github.com/BoltzExchange/boltz-web-app/pull/549/commits/941ab52e9c12b51b22e984beff91671ea4a5a3c4

src/components/ErrorLog.tsx
kilrau: Agree, we can leave it out to make the UI cleaner. Doesn't add any value if my counter is 1 or 1243844.

kilrau:
@michael1011 as just discussed with @dni we want a test to see if clean up after 1 week works

PR #590, opened 2024-05-15
feat: wording successful swap and show amount received

### LINKED ISSUE #584: Show swap amounts in success screen
If I'm looking up a swap in the history (especially liquid with ct) I have no idea which amounts were swapped. Let's simply change the sentence to "You successfully swapped 10000 Liquid sats for 99000 Lightning sats."

image
image

dni:
closes #584

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -54,7 +54,7 @@ const dict = {
             "Paste a bolt11 lightning invoice\n or a Lightning address\nor a LNURL Paylink",
         congrats: "Congratulations!",
         successfully_swapped:
-            "You successfully received {{ amount }} {{ denomination }}.",
+            "You successfully received {{ amount }} {{ denomination }}!",
         timeout_eta: "Timeout ETA",
         pay_invoice: "Swap: {{ id }}",
         pay_swap_404: "Swap not found!",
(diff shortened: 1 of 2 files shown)


src/i18n/i18n.ts
```diff
@@ -427,7 +429,8 @@ const dict = {
         create_and_paste:
             "Pegar una factura Lightning bolt11\n o una dirección Lightning\n o un enlace LNURL Pay",
         congrats: "¡Felicitaciones!",
-        successfully_swapped: "Su intercambio se realizó con éxito!",
+        successfully_swapped:
+            "Has intercambiado con éxito {{ amount }} {{ denomination }}",
```
michael1011, line 433: @kilrau check translations

michael1011 (approved):

dni:
> We want to state both amounts (from / to) with the according layer as I mentioned before "You successfully swapped 10000 Liquid sats for 99000 Lightning sats."
i would not go for having that asset here, you can clearly see the assets and where it was swapped from the asset. also `0.01 BTC btc to 0.009 LN btc` will look quite weird.

i could go for `You successfully swapped {amountSend} {denomination} for {amountReceive} {denomination}`

kilrau (approved):

PR #591, opened 2024-05-16
feat: add `Dockerfile`

dni:
idea kilian and me discussed, can be useful in the future for people to run it themselves or for testing, etc.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,18 @@
-FROM node:20
+FROM node:20 AS builder
+
 WORKDIR /app
-COPY package.json .
-COPY package-lock.json .
+
+COPY package.json package-lock.json .
 RUN npm ci
 COPY . .
 
 ENV NETWORK=mainnet
-ENV DOCKER=1
 
 RUN npm run $NETWORK
 RUN npm run build
-EXPOSE 4183
-CMD ["npm", "run", "serve"]
+
+FROM nginx:alpine AS final
+
+COPY --from=builder /app/dist /usr/share/nginx/html
+
+EXPOSE 80
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,7 +1,7 @@
 ---
 description: >-
-    We encourage our technical users to check the code and run the web app locally
-    from source following the instructions below.
+  We encourage our technical users to check the code and run the web app locally
+  from source following the instructions below.
 cover: .gitbook/assets/boltz-web_app_header.png
 coverY: 0
 ---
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node
```
michael1011, line 1: Set a tag for base images

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+WORKDIR /app
+COPY package.json .
```
michael1011, line 3: The `package-lock.json` too

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN npm i
```
michael1011, line 4:
```suggestion
RUN npm ci
```

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+COPY . .
+
+ENV NETWORK=mainnet
+ENV NODE_ENV=docker
```
michael1011, line 8: Not a valid value https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production#nodejs-the-difference-between-development-and-production

docs/README.md
```diff
@@ -12,6 +12,13 @@ coverY: 0
 
 Make sure to have the latest [Node.js LTS and NPM](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm) installed. We recommend using [nvm](https://github.com/nvm-sh/nvm#install--update-script) to manage npm installs: `nvm install --lts`
 
+## with docker
```
michael1011, line 15:
```suggestion
## With docker
```

vite.config.mjs
```diff
@@ -4,6 +4,8 @@ import mkcert from "vite-plugin-mkcert";
 import { nodePolyfills } from "vite-plugin-node-polyfills";
 import solidPlugin from "vite-plugin-solid";
 
+const isNotDocker = process.env.NODE_ENV !== "docker";
```
michael1011, line 7: Don't abuse `NODE_ENV` for this

vite.config.mjs
```diff
@@ -12,7 +14,7 @@ const commitHash = child
 export default defineConfig({
     plugins: [solidPlugin(), mkcert(), nodePolyfills()],
     server: {
-        https: true,
+        https: isNotDocker,
```
michael1011, line 17: Why no https via Docker?

michael1011 (changes requested):

vite.config.mjs
dni: its better else you need to import the vite certs created inside the docker container or have cert errors. if you wont to expose it it better to run a proxy with ssl in front

michael1011: We are working under the assumption that we are running with HTTPS. Eg clipboard API does not work when you are on HTTP. We need HTTPS

dni: cclipboard works without htrtps, i think its just about the camera feature

michael1011: > Only supported on pages served over HTTPS.
https://stackoverflow.com/a/30810322

HTTPS and localhost

michael1011 (approved):

PR #596, opened 2024-05-22
feat: use trash icon instead of `delete` in SwapList component

dni:
this makes german translations a one liner again
screenshot-1716362226
screenshot-1716362218


kilrau:
A bit hard to click on mobile, button a bit larger with a bit of distance between the buttons would help I think
image

michael1011 (approved):

PR #603, opened 2024-05-28
feat: add Bull Bitcoin logo

dni:


kilrau (approved):
Updated DH link too, currently they have a redirect for .community to .technology but they might let that domain lapse at some point

PR #606, opened 2024-05-28
fix: increase refund QR size

michael1011:
Taproot swaps and especially chain swaps save much more data in the refund file. We need to increase the size of the QR code so that it can be scanned properly.


dni (approved):

kilrau (approved):
Tested chain swaps with old 400px, pretty bad we missed that. New 1200px works.

PR #607, opened 2024-05-29
fix: initialize secp before creating claim transaction

michael1011:
For using Musig, we need the WASM version of secp-zkp to be initialized. That was not the case on page reload because the function that constructs claim transactions was not calling the initializer of secp-zkp.


dni (approved):

kilrau (approved):
Issue reproduced on main, this fixes it.

PR #624, opened 2024-06-26
feat: show cooperative refund error on broadcast fail

michael1011:


src/utils/errors.ts
```diff
@@ -1,6 +1,16 @@
 export const formatError = (message: unknown): string => {
     if (typeof message === "string") {
         return message;
+    } else if (
+        typeof message === "object" &&
```
maybeast, line 5:
```
        const msgObj = message as Record<string, unknown>;
        if (typeof msgObj.message === "string") {
            return msgObj.message;
        }
        if (typeof msgObj.error === "string") {
            return msgObj.error;
        }
```

maybeast (approved):

dni (approved):

PR #631, opened 2024-07-11
feat: improve on `SwapList`

### LINKED ISSUE #609: Fix history line breaks on Mobile
Unsaved Image 2

* Line break before "Created" if it doesn't fit all in one line
* Monospace swap ids

dni:
closes #609 
more room for mobile, remove buttons for simplicity

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
             "version": "1.4.1",
             "dependencies": {
                 "@bitcoinerlab/secp256k1": "^1.1.1",
+                "@fontsource/noto-mono": "^5.0.11",
                 "@fontsource/noto-sans": "^5.0.22",
                 "@scure/base": "^1.1.7",
                 "@solid-primitives/i18n": "^2.1.1",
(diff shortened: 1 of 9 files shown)


dni:
> To make it look aligned, swap id's need to have a fixed length, or at least space. Same with date. Removing the delete button on mobile is good.
> 
> image

the only way we could solve this would be using a monospace font here, should we really include another font just for this? i dont mind it to be different sizes, the layout allows it

screenshot-1721720483

kilrau:
Try the monospace font for swap IDs please too

dni:
screenshot-1721728236

kilrau (commented):
Looking good now. Do you apply it to  the "Refundable" swap list too?

image
image

dni:
yes, same component

michael1011 (changes requested):
When deleting a swap, it navigates to the `/swap/<id>` which immediately shows not found because it has been deleted

michael1011 (approved):

PR #633, opened 2024-07-13
fix: swap box clears address on asset switch

### LINKED ISSUE #611: Swap box clears address on asset switch
Swap box clears address (where it shouldn't) on asset switch:
https://github.com/BoltzExchange/boltz-web-app/assets/19181985/f4ed222f-07b8-41b0-b7d4-d2e87e132a11

Let's not forget tests.

dni:
closes #611

### DIFF
--- a/src/components/AssetSelect.tsx
+++ b/src/components/AssetSelect.tsx
@@ -30,14 +30,15 @@ const SelectAsset = () => {
     const changeAsset = (newAsset: string) => {
         if (isSelected(newAsset)) return;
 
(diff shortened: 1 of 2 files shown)


michael1011 (approved):
Works!

But the same thing should be done for invoices too

dni:
i though maybe not clearing could be dangerous because fee might change and amount too? i could maybe could catch that aswell when i check for the Side Enum

michael1011:
Ah, right

PR #634, opened 2024-07-13
fix: error message was swallowed by JSON.stringify

### LINKED ISSUE #602: bug: log is hiding the real path of the exception
see screenshot.

screenshot-1716811963

dni:
after trying to reproduce #602 i found that it already was fixed, but there was still an issue where the errormessage wasn't found in the logs. issue was `typeof Error` return `object` so the logger treated is as object and tried to JSON.stringify it. that made it to display just `{}`

### DIFF
--- a/src/utils/logs.ts
+++ b/src/utils/logs.ts
@@ -43,6 +43,9 @@ export const deleteOldLogs = async (logsForage: LocalForage) => {
 export const formatLogLine = (message: any[]) =>
     message
         .map((entry: any) => {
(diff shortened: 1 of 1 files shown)


michael1011 (approved):
But this does not fix #602

dni:
i thought you fixed #602 because testing this, i could not reproduce the initial issue, any idea how i did it :D?

michael1011:
Ah, nvmd. You right

PR #635, opened 2024-07-13
fix: error on camera permissions denied

### LINKED ISSUE #600: Throw error when denying camera permissions on mobile
Currently this is the result clicking "Scan QR Code" and then denying camera permissions. The X is hard to find to even unblock the web app again, it should throw a "Camera Permissiosn Denied" error and the web app should continue working and prssing "Scan QR Code" button should throw this error once permissions are denied. 

image

dni:
closes #600

@kilrau translations

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -202,6 +202,7 @@ const dict = {
         reckless_mode_setting: "Reckless Mode",
         reckless_mode_setting_tooltip:
             "Disables prompts to download refund file and other confirmation steps",
+        error_starting_qr_scanner: "Error starting QR scanner",
     },
     de: {
         language: "Deutsch",
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -202,7 +202,8 @@ const dict = {
         reckless_mode_setting: "Reckless Mode",
         reckless_mode_setting_tooltip:
             "Disables prompts to download refund file and other confirmation steps",
-        error_starting_qr_scanner: "Error starting QR scanner",
+        error_starting_qr_scanner:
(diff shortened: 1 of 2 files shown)


src/i18n/i18n.ts
```diff
@@ -201,6 +201,7 @@ const dict = {
         reckless_mode_setting: "Reckless Mode",
         reckless_mode_setting_tooltip:
             "Disables prompts to download refund file and other confirmation steps",
+        error_starting_qr_scanner: "Error starting QR scanner",
```
michael1011, line 204: How about asking to check the permissions?

dni, line 204:
```suggestion
        error_starting_qr_scanner: "Error starting QR scanner, check you camera permissions!",
```

kilrau (changes requested):
Strings added and tested - looking good. What I do not like is that the "Scan QR Code" button disappears after a failed permissions. One has to refresh the site to get a new try. Just keep the button there please.
[Screencast from 07-16-2024 11:39:08 AM.webm](https://github.com/user-attachments/assets/9b476d8b-7d3d-4ee8-9a08-1c98b524dd56)

dni:
> Strings added and tested - looking good. What I do not like is that the "Scan QR Code" button disappears after a failed permission request. One has to refresh the site to get a new try. Just keep the button there please. [Screencast from 07-16-2024 11:39:08 AM.webm](https://github.com/user-attachments/assets/9b476d8b-7d3d-4ee8-9a08-1c98b524dd56)

done

kilrau (approved):
"Scan QR" button doesn't disappear now. Error comes up every time I press it. Behaves as expected.

michael1011 (approved):

PR #636, opened 2024-07-15
fix: prevent pasting same value twice

### LINKED ISSUE #546: Small fixes for localized float separator
Follow-up fix for https://github.com/BoltzExchange/boltz-web-app/pull/538:
- If I have valid amounts set with `.`, but then paste a `,` amount, I end up with this
Screenshot from 2024-03-30 18-46-15
- If amount field is empty and I paste a `,` amount, it converts it to `.`
[Screencast from 03-30-2024 06:50:33 PM.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/445b20cd-b806-4868-8fcd-65ab1530597f)

dni:
closes #546

### DIFF
--- a/src/pages/Create.tsx
+++ b/src/pages/Create.tsx
@@ -148,6 +148,9 @@ const Create = () => {
         }
     };
 
+    const sanitizeInputValue = (value: string) =>
+        value.replace(",", ".").replace(" ", "");
+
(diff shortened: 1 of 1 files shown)


dni:
currently now test for this behavior, but i will add one with this https://github.com/BoltzExchange/boltz-web-app/issues/637

Issue #521, opened 2024-03-20
Add optional 3rd party endpoints for tx broadcasting and reading the chain


kilrau:
We want to enable users to broadcast and read the chain via 3rd party endpoints. But @michael1011 brought up a valid point as where some users might not want their transaction info to be sent to a third party and prefer to still broadcast via Boltz, so we probably want to allow changing endpoints via settings menu but stick with ours by default.

Options in settings menu:
1. [Boltz](https://api.boltz.exchange/swagger)
2. [mempool.space](https://mempool.space/docs/api/rest#post-transaction)
3. [blockstream.info](https://github.com/Blockstream/esplora/blob/master/API.md#post-tx)

CORS might be an issue as always. And we can't really do that before we have DiscountCT is available as our endpoints need to be used for lowball for now.

kilrau:
Crucial for this issue is the case: boltz backend down and make uncooperative claims and refunds work via third party endpoints. Since cooperative path is always preferred we want to handle these similarly to boltz-client:
refunds: retries coop refund every block until timeout block height reached, then goes scriptpath
claims: tries coop 3 times, if these fail goes scriptpath

kilrau:
This is currently blocked by lack of DiscountCT on Liquid mainnet, which we expect to be available in the next months. Then we can implement this.

kilrau:
DiscountCT is available now.

michael1011:
Closed by #802

PR #639, opened 2024-07-15
feat: add blockexplorer setting for links and claiming

dni:
- adds setting to choose a blockexplorer for addresses/txid's
- adds functionality to choose where to broadcast your claim/refund tx. related to https://github.com/BoltzExchange/boltz-web-app/issues/521

### DIFF
--- a/src/components/BlockExplorer.tsx
+++ b/src/components/BlockExplorer.tsx
@@ -2,7 +2,8 @@ import { chooseUrl, config } from "../config";
 import { useGlobalContext } from "../context/Global";
 
 const blockExplorerLink = (asset: string, isTxId: boolean, val: string) => {
-    const basePath = chooseUrl(config.assets[asset].blockExplorerUrl);
+    const url = config.assets[asset].blockExplorerUrls?.mempool;
+    const basePath = chooseUrl(config.assets[asset].blockExplorerUrls?.mempool);
     return `${basePath}/${isTxId ? "tx" : "address"}/${val}`;
 };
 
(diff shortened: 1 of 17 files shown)


kilrau:
Gave this a quick glance, should work but need to think about how we handle cases like "Liquid Lowball fees not available via external services"

michael1011:
> but need to think about how we handle cases like "Liquid Lowball fees not available via external services"

You just switch to 0.1 sat/vbyte and only allow switching to a non-lowball endpoint when there are no pending reverse or chain swaps

kilrau:
> only allow switching to a non-lowball endpoint when there are no pending reverse or chain swaps

Better allow switching any time but only apply to swaps created after enabling.

So you'd be fine with putting "Alternative endpoints do not feature low Liquid fees and will use the standard 0.1 sat/vbyte" in the `?` part and that's it? @michael1011

michael1011:
> So you'd be fine with putting "Alternative endpoints do not feature low Liquid fees and will use the standard 0.1 sat/vbyte" in the ? part and that's it?

Yes

kilrau:
> > only allow switching to a non-lowball endpoint when there are no pending reverse or chain swaps
> 
> Better allow switching any time but only apply to swaps created after enabling.
> 
> So you'd be fine with putting "Alternative endpoints do not feature low Liquid fees and will use the standard 0.1 sat/vbyte" in the `?` part and that's it? @michael1011

Ayt let's do it this way then. Fees currently don't update when I choose a non-Boltz "Broadcaster".

The thing I am not happy with yet is the button style. We are abusing a toggle (which is meant to hold 2 very simple states) for a dropdown now. Should be similar style than asset selector. A toggle just doesn't work for writing out longer words, let alone three.

image

PR #643, opened 2024-07-19
fix: Lightning Nodes i18, remove link from headline

kilrau:
- node*s* (plural)
- remove link in the headline, with public explorers it can only link to one node and it was broken on testnet

### DIFF
--- a/src/pages/Hero.tsx
+++ b/src/pages/Hero.tsx
@@ -27,10 +27,6 @@ export const Hero = () => {
     const { hideHero, setHideHero, t, denomination, separator } =
         useGlobalContext();
 
-    const openNodeInfo = async () => {
-        window.open(`https://api.boltz.exchange/v2/nodes/`, "_blank");
-    };
-
     const formatStatsAmount = (
         value: number,
         denom: Denomination = Denomination.Sat,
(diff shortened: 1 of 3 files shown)


src/pages/Hero.tsx
```diff
@@ -27,10 +25,6 @@ export const Hero = () => {
     const { hideHero, setHideHero, t, denomination, separator } =
         useGlobalContext();
 
-    const openNodeInfo = async () => {
-        window.open(`${config.lightningExplorerUrl}/${nodePubkey()}`, "_blank");
```
michael1011, line 31: This was the only place we used `lightningExplorerUrl`. Also remove it from `config.ts`

src/pages/Hero.tsx
```diff
@@ -43,9 +37,6 @@ export const Hero = () => {
                 getNodeStats(BTC),
             ]);
             log.debug("node", nodesRes);
```
michael1011, line 39: If you remove this log statement, we don't need to fetch `nodeRes` at all

michael1011 (changes requested):

src/pages/Hero.tsx
kilrau: Done

src/pages/Hero.tsx
kilrau: Done

michael1011 (approved):

PR #649, opened 2024-07-23
feat: add 0conf setting

dni:
wip

### DIFF
--- a/src/components/settings/ZeroconfSetting.tsx
+++ b/src/components/settings/ZeroconfSetting.tsx
@@ -1,0 +1,24 @@
+import { useGlobalContext } from "../../context/Global";
+
+const ZeroconfSetting = () => {
+    const { zeroconf, setZeroconf, t } = useGlobalContext();
+
+    const toggleZeroconf = (evt: MouseEvent) => {
+        setZeroconf(!zeroconf());
+        evt.stopPropagation();
+    };
+
+    return (
+        <>
+            <div
+                class="zeroconf toggle"
+                title={t("enable_zeroconf_tooltip")}
+                onClick={toggleZeroconf}>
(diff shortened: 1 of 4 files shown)


src/components/settings/ZeroconfSetting.tsx
```diff
@@ -0,0 +1,24 @@
+import { useGlobalContext } from "../../context/Global";
+
+const ZeroconfSetting = () => {
```
michael1011, line 3: This is the fourth true/false switch. Let's unify those with one base component

src/i18n/i18n.ts
```diff
@@ -204,6 +204,10 @@ const dict = {
             "Disables prompts to download refund file and other confirmation steps",
         error_starting_qr_scanner:
             "Couldn't access camera, please check permissions!",
+        zeroconf: "Zeroconf",
+        zeroconf_tooltip: "Enable or disable zeroconf",
+        enable_zeroconf: "Enable Zeroconf",
+        enable_zeroconf_tooltip: "Enable or disable zeroconf larger desc",
```
michael1011, line 210: ?

src/i18n/i18n.ts
```diff
@@ -204,6 +204,10 @@ const dict = {
             "Disables prompts to download refund file and other confirmation steps",
         error_starting_qr_scanner:
             "Couldn't access camera, please check permissions!",
+        zeroconf: "Zeroconf",
```
michael1011, line 207: Call it `0-conf` in labels

michael1011 (changes requested):
Where is the implementation of that feature?

dni:
still wip, i only through bare bones out there

PR #658, opened 2024-08-15
WebSocket fallback endpoint

michael1011:


src/components/SwapChecker.tsx
```diff
@@ -41,45 +42,18 @@ class BoltzWebSocket {
...
-                    await this.swapClaimLock.acquire(() =>
-                        this.claimSwap(status.id, status),
-                    );
-                }
-            }
-        };
+        log.debug("Opening WebSocket");
+        this.openWebSocket(`${this.url}/v2/ws`).catch(() => {
+            log.debug("Opening fallback WebSocket");
+            this.openWebSocket(this.wsFallback).then();
```
maybeast, line 55: when fallback is not defined (testnet/mainnet) it'll attempt to connect once and throw an error

src/config.ts
```diff
@@ -45,7 +44,9 @@ type Url = {
 };
 
 export type Config = {
-    apiUrl?: Url;
+    // The wsFallback is used on regtest when the backend is run without nginx
```
maybeast, line 47:
```suggestion
    // The wsFallback is used on regtest when the backend is being run without nginx
```

maybeast (changes requested):

PR #677, opened 2024-10-03
feat: show swap ID after uploading file

### LINKED ISSUE #676: Show SwapID in Refund Page
Had some support cases recently where the only thing user's had was the refund QR code. To see what's going on, I need the swap's id. Many users are struggling to decode the QR code and copy the swap id from the json-encoded string. Easier would be if I could just tell them to upload the qr in the refund section of our web page and send me a screenshot.

Simply change the title "Refund a swap" to "Refund Swap ID" once the qr code/json file is present:
image
image

michael1011:
Closes #676

Needs translations


kilrau (approved):
LGTM
image

PR #704, opened 2024-10-21
fix: add timeout to BIP-353 DNS lookup

### LINKED ISSUE #703: Web App stuck after fetching invoice via LNURL
On current master mainnet I can reliably reproduce web app getting stuck after fetching the invoice via lnurl from my alby account. Doesn't happen on current mainnet deployment so I am guessing it's related to recent bolt12 merges.

image

michael1011:
Closes #703

@kilrau check translations and if it fixes the issue on the weird mobile connection

### DIFF
--- a/src/components/CreateButton.tsx
+++ b/src/components/CreateButton.tsx
@@ -160,7 +160,7 @@ export const CreateButton = () => {
                     (() => {
                         return new Promise<string>(async (resolve, reject) => {
                             const timeout = setTimeout(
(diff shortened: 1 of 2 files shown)


kilrau (approved):
works now even though 5s feels long, minor wording fix is pushed

michael1011:
> works now even though 5s feels long

Extraordinary situation in which DoH is blocked and we gotta consider slow LNURL servers, nodes and client internet connections

PR #706, opened 2024-10-22
feat: validate invoices fetched for offers

michael1011:


src/utils/invoice.ts
```diff
@@ -248,3 +248,48 @@ export const isBolt12Offer = (offer: string) => {
...
+
+        for (const signer of expectedSigners) {
+            if (signer.length !== invoiceSigner.length) {
+                continue;
+            }
+
+            if (signer.every((b, i) => b === invoiceSigner[i])) {
+                return true;
+            }
+            console.log("no match");
```
maybeast, line 288: remove or add more context. In the current state it's pretty useless.

src/utils/invoice.ts
```diff
@@ -248,3 +248,48 @@ export const isBolt12Offer = (offer: string) => {
...
+        expectedSigners.push(of.signing_pubkey);
+    }
+
+    for (const path of of.paths) {
+        const hops = path.hops;
+        if (hops.length > 0) {
+            expectedSigners.push(hops[hops.length - 1].pubkey);
+        }
+
+        for (const hop of hops) {
```
maybeast, line 266:
```suggestion
        hops.forEach(hop => hop.free());
```

maybeast (approved):

kilrau:
2/10 of my testnet bip353 resolve tests needed too long to wake up my phoenix app. Happened when I restarted my phone to make sure Phoenix is not running in the background already. But the notification literally came in a second or two after web app timed out with "failed to fetch". I suggest to bump the timeout for about 3s.

kilrau (changes requested):
Refund qr code generation fails (json works). Also we should throw a tool tip error if that happens and block the jump to the next screen to not let users create swaps without refund file.

image

kilrau (approved):
Moved above comments into separate issues. Invoice <> Offer validation works.

PR #734, opened 2024-11-11
feat: Submarine Swap preimage copy button

### LINKED ISSUE #732: Submarine Swaps: Include "Payment Proof" copy button in success page
This has to be stored in local storage from the response of `/swap/submarine/{id}/claim`. Once the swap was successful, it can't be pulled anymore. I would not show the preimage as text string as it's long and ugly but as "Payment Proof *copy icon*" button. 

I handed out the preimage a ton recently in support cases for folks to proof an invoice was paid at a merchant (seems to always have convinced the merchant so far) and the feature request exists [since a while](https://discord.com/channels/547454030801272832/714342346568892487/1164762317439914085).

michael1011:
Closes #732 

Depends on https://github.com/BoltzExchange/boltz-backend/pull/720

@kilrau needs translations

### DIFF
--- a/src/status/TransactionClaimed.tsx
+++ b/src/status/TransactionClaimed.tsx
@@ -1,13 +1,19 @@
 import { useNavigate } from "@solidjs/router";
 import { BigNumber } from "bignumber.js";
-import { Show, createEffect, createSignal } from "solid-js";
+import log from "loglevel";
+import { Show, createEffect, createResource, createSignal } from "solid-js";
 
+import CopyButton from "../components/CopyButton";
 import LoadingSpinner from "../components/LoadingSpinner";
 import { RBTC } from "../consts/Assets";
 import { SwapType } from "../consts/Enums";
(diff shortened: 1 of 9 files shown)


src/status/TransactionClaimed.tsx
```diff
@@ -23,12 +26,28 @@ const Broadcasting = () => {
...
+            return undefined;
+        }
+
+        if (submarine.preimage !== undefined) {
+            return submarine.preimage;
+        }
+
+        const res = await getSubmarinePreimage(submarine.id);
+        submarine.preimage = res.preimage;
+        await setSwapStorage(submarine);
```
jackstar12, line 47: We could validate the preimage here aswell, but boltz doesnt really have any incentive to lie about it so ig its fine.

michael1011: Still, you are right. Let's do a check

kilrau (approved):
tACK
image

PR #745, opened 2024-11-25
feat: add chatwoot, replace discord as main help link

kilrau:


index.html
```diff
@@ -112,5 +112,21 @@
         <script
             src="/node_modules/iframe-resizer/js/iframeResizer.contentWindow.js"
             type="module"></script>
+        <script>
+            (function(d,t) {
+                var BASE_URL="https://support.boltz.exchange";
```
michael1011, line 117: Should be configurable

michael1011 (changes requested):
The Chatwoot script should be loaded programmatically in the entrypoint of the Solid application *if* its token is set in the `.env` file

michael1011 (changes requested):
GitHub posted the review twice

kilrau:
Can't approve as it's my own PR, but LGTM

src/chatwoot.ts
```diff
@@ -0,0 +1,55 @@
...
+            };
+        }
+    });
+
+    const { swap } = usePayContext();
+
+    createEffect(() => {
+        if (swap() !== null) {
+            window.$chatwoot.setCustomAttributes({
+                swapId: swap().id,
```
michael1011, line 49: Is not propagated to the Chatwoot UI

michael1011 (changes requested):

src/chatwoot.ts
```diff
@@ -0,0 +1,55 @@
...
+            setCustomAttributes: (attributes: { swapId: string }) => void;
+        };
+    }
+}
+
+export default function Chatwoot() {
+    onMount(() => {
+        const token = import.meta.env.VITE_CHATWOOT_TOKEN;
+        const url = config.chatwootUrl;
+        if (token !== undefined && url !== undefined) {
```
michael1011, line 24: Let's also do a `log.warn` in case it's not set

PR #761, opened 2024-12-03
refactor: make RSK log scan endpoint optional

michael1011:


kilrau (approved):
Working like a charm :pray:

Issue #747, opened 2024-11-25
Refund page: Split up into two sections (UTXO/EVM)


kilrau:
1. In the initial state of our refund page, the "connect wallet" button exists to be able to connect a Rootstock wallet and scan for refundable swaps.
![image](https://github.com/user-attachments/assets/274975dc-beb6-44c9-b333-c57cb5f4565d)
BUT that's just super confusing for users who never used or don't plan to use Rootstock or are not thinking about Rootstock in this moment. It feels like the "continue" button - that I have to connect a wallet to do a Liquid refund. 
![image](https://github.com/user-attachments/assets/202d127a-7b10-4487-bc9a-ef680ce15487)

Also it is confusing for users that don't know that a Rootstock wallet can be scanned for refundable swaps. We should make it clear that refund section has two parts: file upload is for swaps originating from UTXO chains like Liquid, the wallet connect button for swaps originating from EVM chains like Rootstock. I propose to use the following descriptions headlining each:
`For refunds originating from UTXO chains like Liquid, Upload your refund file and reclaim your locked funds.`
`For refunds originating from EVM chains like Rootstock, connect your wallet below to scan for refundable swaps.`

Maybe with a separator line in between them.

2. When a wallet is connected, fully scanned and doesn't have refundable swaps we should say that in a button below `No refundable swaps found` otherwise the button should say `Refund Swap` to open the swap for refund. I had support cases where users just didn't understand to enter the displayed swap via the `View` button. Maybe we should call the `View` button `Refund` in this state.
![image](https://github.com/user-attachments/assets/2452e15f-2129-41b1-9804-7fe02aaec03c)

Yes, this creates two entries into refunding a swap: one big button on the bottom and one small `View/Refund` button for each swap in the list and that is wanted because there might be multiple refundable swaps found in a rootstock wallet the same way utxo chains can have multiple refundable swaps in local storage and users can enter them individually by pressing `View/Refund`. The big `Refund button` below just always enters the first refundable swap and exists for the sake of users not missing the button in the 99% case where there is only one refundable swap.

3. If a non-Rootstock refund file is uploaded this is the state. I would completely remove the rootstock part for this state. 
![image](https://github.com/user-attachments/assets/8a2f53c7-5935-48ec-9d8d-c81cc698f860)

4. The same way I would remove the Refund file upload section when a refundable rootstock swap is displayed. 

So /refund offers two ways to start out: utxo and evm refund but when one is active, it only shows one.

[...]

PR #785, opened 2025-01-03
refactor: split up refund page

jackstar12:
closes: #747 

Also removes the link to the existing swap upon refund file upload if its found within browser history - its simply handled the same as all other refund files.


kilrau (changes requested):
Minor wording discrepancies with Figma in many places:
- Landing title should be "Refund Swap" not "Refund a swap", please check all figma screens. Add fullstops etc.
- Bitcoin/Liquid should have address field and refund button on empty state already and not have it appear automagically after selecting refund file (as we have it rn) 
- I forgot to include this error state in the wireframe, but here the sentence should say "Please switch your wallet network to Rootstock"
image


Other than that awesome improvement, I really like this:

image

jackstar12:
@kilrau addressed all except the last one - will be in a follow up pr.

src/pages/RefundExternal.tsx
```diff
@@ -0,0 +1,255 @@
...
+            setLogRefundableSwaps(logRefundableSwaps().concat(value.events));
+        }
+
+        setRefundScanProgress(undefined);
+    });
+
+    return (
+        <Show
+            when={import.meta.env.VITE_RSK_LOG_SCAN_ENDPOINT !== undefined}
+            fallback={<p>{t("rsk_log_endpoint_not_available")}</p>}>
```
michael1011, line 190: Let's not show this in the UI, because there is nothing that can be done about it when you open the page. A  `log.warn` should be enough

src/index.tsx
```diff
@@ -105,6 +106,8 @@ const cleanup = render(
             <Route path="/swap/refund/:id" component={RefundStep} />
             <Route path="/error" component={() => <Error />} />
             <Route path="/refund" component={Refund} />
+            <Route path="/refund/external" component={RefundExternal} />
+            <Route path="/refund/external/:type" component={RefundExternal} />
```
michael1011, line 110: Not sure if we should have `/refund/external` as a route. `/refund/external` and `/refund/external/<default>` are the same thing this way and it's kinda weird

michael1011 (commented):
utACK. Let me play around on regtest a little

michael1011 (changes requested):
image
This is super weird

---

image

My "no lockup transaction" logic is gone

---

Even when there is a transaction to be refunded it is not fetched
```
refund failed TypeError: Cannot read properties of undefined (reading 'hex')
    at refund (refund.ts:197:29)
    at async refundAction (RefundButton.tsx:167:25)
```

jackstar12:
> This is super weird

Not sure about it either. Was @kilrau feedback

> My "no lockup transaction" logic is gone

It's because of https://github.com/BoltzExchange/boltz-web-app/pull/785/commits/c7bb7d36c938cc114d4dab1628969441b2952834. The lockupTransaction resource isn't reactive, so on initial render it doesnt fetch.

src/index.tsx
jackstar12: Otherwise, we would have to naviate to `/refund/external/btc` in the normal refund page. 
I like it more this way, since the default value is within the `RefundExternal` component

kilrau:
> > This is super weird
> 
> Not sure about it either. Was @kilrau feedback
> 
> >

It drew this state in Figma ser, it should only have one button obviously
image

kilrau (changes requested):
> > This is super weird
> 
> Not sure about it either. Was @kilrau feedback
> 
> >

It drew this state in Figma ser, it should only have one button obviously
image

jackstar12:
I think its fine as-is where the address button only appears once a valid refund file is present - why display the address input when its always disabled in that case anyways?

kilrau:
A refund address is always needed no matter what for refunds so the element has to be there from the beginning. Vamos!

jackstar12:
> A refund address is always needed no matter what for Bitcoin/Liquid refunds so the element has to be there from the beginning. Vamos!

But only once you've uploaded your refund file - one thing at a time

kilrau:
> > A refund address is always needed no matter what for Bitcoin/Liquid refunds so the element has to be there from the beginning. Vamos!
> 
> But only once you've uploaded your refund file - one thing at a time

still not negotiable

jackstar12:
Its a bit of a pain since refund button is decoupled from refund json upload stuff.

Why not keep it like it is now? it makes perfect sense to me to only display refund button  / address input once the file is there?

kilrau (changes requested):
1. Let me enter an address in this state!!!
image
2. The address field shouldn't disappear in this state
image

EDIT: some of these error messages (in the button) are missing too
image

kilrau:
and let's fix this CI :muscle:

jackstar12:
> Let me enter an address in this state!!!

What about address validation in that case?

kilrau:
> > Let me enter an address in this state!!!
> 
> What about address validation in that case?

disable it, enable it once a valid refund file was parsed

kilrau:
Created follow up issue for my review: https://github.com/BoltzExchange/boltz-web-app/issues/793  So here only CI is missing.

michael1011 (approved):
tACK

src/components/RefundButton.tsx
```diff
@@ -295,7 +296,11 @@ const RefundButton = (props: {
                 </Show>
             }>
             <Switch>
-                <Match when={lockupTransaction.state === "ready"}>
+                <Match
+                    when={
+                        lockupTransaction.state === "ready" ||
+                        lockupTransaction.state == "unresolved"
```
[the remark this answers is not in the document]
jackstar12: Its the initial state before it has been fetched for the first time.

Since we know pass the swap to the resource as a source signal it only fetches once populated (and refetches when it changes)

Issue #793, opened 2025-01-09
Refund Page Follow Up


kilrau:
1. Let me enter an address in this state, disable address validation and enable it as soon as a valid refund file was parsed
![image](https://github.com/user-attachments/assets/a7f94871-e509-4ce3-94f1-ddcb4ca07333)
2. The address field shouldn't disappear in this state
![image](https://github.com/user-attachments/assets/00bd5f90-c5c5-4a3e-93ed-52eb937ad78a)
3. EDIT: this error state should be there too (just added it to figma)
![image](https://github.com/user-attachments/assets/4219685a-20d2-45d4-bbf2-714c0d34fdd1)

PR #796, opened 2025-01-11
refactor: refund address input

jackstar12:
closes: #793

### DIFF
--- a/src/components/RefundButton.tsx
+++ b/src/components/RefundButton.tsx
@@ -109,7 +109,7 @@ export const RefundEvm = (props: {
     );
 };
 
-const RefundButton = (props: {
+export const RefundBtc = (props: {
     swap: Accessor<SubmarineSwap | ChainSwap>;
     setRefundTxId?: Setter<string>;
     buttonOverride?: string;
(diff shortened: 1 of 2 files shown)


kilrau (changes requested):
Bullet points from issue #793:
1.+2. tACK
3. Error state saying "Enter Address" in the button is missing:
image
Should show "Enter address" in the button on the regular refund from local storage page too:
image

+ CI

jackstar12:
> Error state saying "Enter Address" in the button is missing:

So we wanna tell the user in 3 different places right next to each other to enter an address? Why?

kilrau (approved):

src/components/RefundButton.tsx
```diff
@@ -234,6 +226,68 @@ const RefundButton = (props: {
         return transactionToRefund;
     });
 
+    const message = () =>
+        props.buttonOverride ??
+        (lockupTransaction.state == "errored"
+            ? t("no_lockup_transaction")
+            : valid() || !refundAddress() || !props.swap()
+              ? t("refund")
+              : t("invalid_address", { asset: props.swap()?.assetSend }));
```
michael1011, line 235: Let's not nullish coalesce a double ternary. Now that we have a function for it, we should make it readable

michael1011 (changes requested):
E2E tests are failing

When you enter an address before uploading a file, the input field is reset which is why the E2E tests are failing

michael1011 (approved):

PR #805, opened 2025-01-29
Add labels to denomination buttons

kiocosta:
Add labels to denomination buttons as a way of increasing clarity on the meaning of each symbol.

#### Desktop / Tablet
image

#### Mobile
Aligned items on center to prevent elements from overlapping on small screens

image

### DIFF
--- a/src/components/settings/Denomination.tsx
+++ b/src/components/settings/Denomination.tsx
@@ -1,7 +1,9 @@
 import btcSvg from "../../assets/btc.svg";
 import satSvg from "../../assets/sat.svg";
+import { BTC } from "../../consts/Assets";
 import { Denomination as Denoms } from "../../consts/Enums";
 import { useGlobalContext } from "../../context/Global";
+import { formatDenomination } from "../../utils/denomination";
 
 const Denomination = () => {
     const { denomination, setDenomination, t } = useGlobalContext();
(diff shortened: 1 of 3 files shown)


kilrau:
I like it. Other opinions?

michael1011 (commented):
On desktop it makes sense. Not sure about mobile though

kilrau:
Hmm it should fit left-aligned even on an iPhone SE screen, could you test and fix that? @kiocosta 
image

kiocosta:
@kilrau It does fit, but I noticed the sats icon may jump to another line depending on the value inserted, but I will try to come up with a solution other than aligning center

image

kiocosta:
@kilrau @michael1011 what about this one?

### Mobile
image

I tried to keep denomination icons aligned on fees (as they were in desktop version), but placed them on the left to also align with the other inputs. Desktop version is still the same.

kilrau:
Honestly, it's too much like this. It's a subtle config you probably only change once.

Two approaches how to continue for mobile:
1. Stick with symbol only to make things fit
2. Move btc/sats to settings menu

I tend to prefer 1 as on mobile amounts are often typed by hand and the switch is actually important (when pasting the btc/sats switch triggers smartomagically depending on the pasted amount)

kiocosta:
> Honestly, it's too much like this. It's a subtle config you probably only change once.
> 
> Two approaches how to continue for mobile:
> 
> 1. Stick with symbol only to make things fit
> 2. Move btc/sats to settings menu
> 
> I tend to prefer 1 as on mobile amounts are often typed by hand and the switch is actually important (when pasting the btc/sats switch triggers smartomagically depending on the pasted amount but pasting is more of a desktop thing)

Got it! I have reverted the mobile version to the original toggle and kept the buttons with label for the desktop version.

michael1011 (approved):

Issue #809, opened 2025-01-31
Refund: use 3rd party blockexplorer endpoints to double check if a lockup address has funds


kilrau:
Another annoying side effect of https://github.com/BoltzExchange/boltz-backend/issues/803 is that users can't refund these swaps via Boltz Web App as our backend is not monitoring lockup addresses after swap expiration (backend swap status `"swap.expired", "failureReason": "onchain HTLC timed out"`).

Adding scanning for all lockup addresses ever created on backend side is ... difficult, so we want Boltz Web App itself to check an address for funds on demand in the refund flow and thus enable refunding for lockup addresses that received funds after swap expiration.

Let's have redundancy by having a fail over.
1. https://github.com/Blockstream/esplora/blob/master/API.md#addresses (Bitcoin+Liquid - endpoints in beginning of doc)
2. https://mempool.space/docs/api/rest#get-address (Bitcoin), https://liquid.network/docs/api/rest#get-address (Liquid)

kilrau:
This is especially crucial for swaps that immediately fail because of short invoice expiries (mostly Strike's 59s expiry invoices) as the user is busy sending funds while the swap already failed.

kiocosta:
working on this one

kilrau:
Awesome!! It's a pretty critical issue, several users needed help with manual refunds these days because of this.

PR #811, opened 2025-02-02
fix: Don't show swaps as 'refundable' when there was no lockup tx

### LINKED ISSUE #806: Don't show swaps as "refundable" in /refund when there was no lockup tx
As a side effect of us (as of [recently](https://github.com/BoltzExchange/boltz-backend/pull/762)) setting swaps with expired invoices to status `invoice.failedToPay`, these are now all showing as "Refundable" in our refund page. Previously status `invoice.failedToPay` would only apply if there were funds locked up, we actually tried to pay the invoice and that failed.
Image

Now, status `invoice.failedToPay` also applies to swaps with expired invoices, no matter if there ever were funds locked up or not. Swaps that don't have a lockup tx/funds locked (swap was created by money never moved) should NOT show here as "refundable":
Image

Consequently, WebApp should query swap endpoints to see if there is a lockup tx each and only show the ones that *do* have a lockup tx:
https://api.boltz.exchange/swagger#/Submarine/get_swap_submarine__id__transaction
https://api.boltz.exchange/swagger#/Chain%20Swap/get_swap_chain__id__transactions

kiocosta:
Closes #806

I have tested this by creating a few swaps and:
- letting them expire without paying (these were not displayed);
- paying them after they've expired (these were displayed);
- paying an incorrect amount (these were displayed);

### DIFF
--- a/src/pages/Refund.tsx
+++ b/src/pages/Refund.tsx
@@ -27,17 +27,17 @@ const Refund = () => {
             setRefundableSwaps(refundableSwaps().concat(swap));
         };
 
-        const swapsToRefund = (await getSwaps())
+        const allSwaps = await getSwaps();
+
+        const swapsToRefund = allSwaps
             .filter(refundSwapsSanityFilter)
-            .filter((swap) =>
-                [
-                    swapStatusFailed.InvoiceFailedToPay,
-                    swapStatusFailed.TransactionLockupFailed,
-                ].includes(swap.status),
+            .filter(
+                (swap) =>
+                    swapStatusFailed.TransactionLockupFailed === swap.status,
             );
         setRefundableSwaps(swapsToRefund);
 
-        void (await getSwaps())
+        void allSwaps
             .filter(refundSwapsSanityFilter)
             .filter(
                 (swap) =>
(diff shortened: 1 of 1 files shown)


kilrau (approved):
Sorry for the late review, many of us are travelling. tACK from my side (no code review)

I tested the following 4 scenarios:
1. paying correctly and letting a swap fail (e.g.`invoice.failedToPay`): displayed ✅
2. paying an incorrect amount: displayed ✅
3. letting a swap expire without paying: not displayed ✅
4. With "paying them after they've expired (these were displayed);" you made it sound like you are already did https://github.com/BoltzExchange/boltz-web-app/issues/809 but that's not the case. All good, let's handle #809 separately. ✅

src/pages/Refund.tsx
```diff
@@ -27,17 +27,18 @@ const Refund = () => {
...
+        const swapsToRefund = allSwaps
             .filter(refundSwapsSanityFilter)
             .filter((swap) =>
-                [
-                    swapStatusFailed.InvoiceFailedToPay,
-                    swapStatusFailed.TransactionLockupFailed,
-                ].includes(swap.status),
+                [swapStatusFailed.TransactionLockupFailed].includes(
+                    swap.status,
+                ),
```
michael1011, line 37:
```suggestion
                swapStatusFailed.TransactionLockupFailed === swap.status,
```

michael1011 (changes requested):

src/pages/Refund.tsx
kiocosta: @michael1011 done!

PR #820, opened 2025-02-14
feat: swap rescue rescan

### LINKED ISSUE #91: Boltz Rescue Key: Recover all swaps using a single key file

michael1011:
Closes #91

Depends on https://github.com/BoltzExchange/boltz-backend/pull/820


kilrau (commented):
Concept ACK

- Forcing download of the recovery file in a fresh session for the first swap (no matter which) and then no more prompts afterwards - good
- Recovery file contains one private key and nothing else - good
- Let's add "DO-NOT-SHARE" to the JSON filename?
- Can we add a "Boltz Recovery File - DO NOT SHARE" title into the QR, ideally even a Boltz Logo in the middle of the QR Code so that users can easily identify the file in their phone's gallery?
- Missing: reupload to verify download
- I'll throw up a wireframe for the recovery file download and reupload pages to have a more concrete idea of wording and how this should look like

kilrau:
Finally got around to throw up a wireframe: https://www.figma.com/design/2tCQYrEwE45lxrAT37tBmv/Boltz-Seed-File?node-id=0-1&t=Zm6RFYOD5Me6h1VW-1

Re wording, I was going through the entire recovery/seed/master/refund/emergency/rescue word salad and for me "Boltz Rescue Key" came out top. Combines a sense of "that's important" with "I only need this once". But definitely open for suggestions.

From my comment above still open:
- [ ] Add "DO-NOT-SHARE" to the JSON filename
- [ ] Add "Boltz Rescue Key - DO NOT SHARE" title into the QR, ideally even a Boltz Logo in the middle of the QR Code so that users can easily identify the QR code in their phone's gallery

michael1011:
> Add "Boltz Rescue Key - DO NOT SHARE" title into the QR, ideally even a Boltz Logo in the middle of the QR Code so that users can easily identify the QR code in their phone's gallery

Neither can be done with our current QR code library

kilrau (changes requested):
- [ ] json should be named `boltz-rescue-key-DO-NOT-SHARE.json`
- [ ] `Verify Existing Key` -> `Verify Boltz Rescue Key`, otherwise the download+verify flow is weird if it talks about an "existing" key in the second step. See wireframe.
- [ ] Here the grey button is missing, we want to avoid automagically appearing buttons:
image
- [ ] Success page missing (wireframe)


One more thing which I didn't cover in my wireframe and still not sure what to do: what do we do with the current Refund menu. How about I throw up the pages for uploading the Boltz Rescue Key as new default (if it's not present in browser) with some button linking to the existing single-use refund file flow for backward compatibility? Any other ideas? @michael1011

michael1011:
>  Verify Existing Key -> Verify Boltz Rescue Key, otherwise the download+verify flow is weird if it talks about an "existing" key in the second step. See wireframe.

You need to update the strings for translations. Feel free to reword there

>  Here the grey button is missing, we want to avoid automagically appearing buttons:

There is no button because it proceeds with the logic after the upload

>  Success page missing (wireframe)

Why do we need a success page?

> One more thing which I didn't cover in my wireframe and still not sure what to do: what do we do with the current Refund menu. How about I throw up the pages for uploading the Boltz Rescue Key as new default (if it's not present in browser) with some button linking to the existing single-use refund file flow for backward compatibility? Any other ideas?

Why do we need to change the current flow? External refunds handle the recovery file and single use ones

kilrau:
> You need to update the strings for translations. Feel free to reword there

Will do.

> There is no button because it proceeds with the logic after the upload

I am aware what it does, but magic things happening like a page jump are bad UX and also not consistent with our regular refund file upload which has an always visible button below: 
Screenshot from 2025-02-27 11-59-57. I have to insist to stay consistent here.

> Why do we need a success page?

To signify that all is good. Not saying anything leaves the user in a "why did I have to do that now" state. Also, as this is only applying to first time users are e.g. Tor Browser users the additional click for the sake of clarity is justified imo.

> Why do we need to change the current flow? External refunds handle the recovery file and single use ones

Yes but it's confusing to have two things or rather three (Rescue Key, Single use refund files, local storage). Anyhow, for another PR.

kilrau:
Also what just comes to mind. Once downloaded and verified we currently don't offer a way to download again. Or a new one. If I accidentally deleted my rescue key, that should be possible. Wdyt?

michael1011:
> I am aware what it does, but magic things happening like a page jump are bad UX and also not consistent with our regular refund file upload which has an always visible button below:

That refund flow sends a transaction that cannot be reversed. It makes sense to ask for confirmation. This here is asking for a file input. You give it a file input and things happen. How is that bad ux?

> To signify that all is good. Not saying anything leaves the user in a "why did I have to do that now" state. Also, as this is only applying to first time users are e.g. Tor Browser users the additional click for the sake of clarity is justified imo.

It working tells you it's all good. If there is confusion about why that was neccessary, we can adjust later. This is not going to help anyone (also, it's not a browser session)

image

> Yes but it's confusing to have two things or rather three (Rescue Key, Single use refund files, local storage). Anyhow, for another PR.

How is it asking for a file, you giving it a file and things working confusing? If anything, adding a section makes it way worse. What happens if you upload a file in the wrong tab? Do you show an error to use the other tab? If we handle the scenario of uploading in the wrong tab, why don't we switch we automatically? If we switch automatically, why don't use we use the same tab?

Also why is uploading a file and things magically happening fine here? Why don't we have another button to confirm the file? Because it asking for a file and you giving it a file is fine.

> Also what just comes to mind. Once downloaded and verified we currently don't offer a way to download the key or a new one again. If I accidentally deleted my rescue key and am aware of that, this should be possible.

Let's just put it in the settings

kilrau:
Why have this id in the url?
image

michael1011:
> Why have this id in the url?

So that you can refresh without it loosing state

> EDIT: also I think we can remove /existing from the backup/verify/existing url path. No need for three levels here

Fine

> Also current behavior is that /backup/verify takes me back to the download which is confusing

Yes, editing the URL, and with that the state of the application, causes undefined behaviour

> When uploading an existing rescue key on a clean browser storage/db I can reliably reproduce web app jumping back to the initial /swap state instead of creating the swap and jumping to /swap/ID

Has to be that way. The swap was created with the freshly generated key but you imported something else. The created swap is not covered by the rescue key you imported

kilrau:
Looks like the legacy refund menu is bricked. We agreed leaving it 1:1 as-is no?

image

michael1011:
> Looks like the legacy refund menu is bricked. We agreed leaving it 1:1 as-is no?

If you don't have a RSK RPC endpoint configured in your env, the Rootstock option does not show. The public node you have configured in your MetaMask does not support scanning logs

kilrau:
> Has to be that way. The swap was created with a freshly generated key but you imported something else. The created swap is not covered by the rescue key you imported

I see, you are creating the swap already before the verify existing/download new page comes up. We can do this as follow up but I can see our tor/incognito users being very annoyed by that behavior as they have to enter everything twice all the time. So can't we just create a second swap with the uploaded key under the hood and then move on? Should be possible but I understand that's a piece of work.

michael1011:
> We can do this as follow up but I can see our tor/incognito users being very annoyed by that behavior as they have to enter everything twice all the time. So can't we just create a second swap with the uploaded key under the hood and then move on? Should be possible

Not possible for submarine swaps and creates a billion edge cases. What we can do defer creating the swap until the backup is done. In case creation fails, send them back to the swap box but with all values still there

kilrau:
> What we can do defer creating the swap until the backup is done. In case creation fails, send them back to the swap box but with all values still there

Sounds good!

src/pages/RefundRescue.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { RescuableSwap } from "../utils/boltzClient";
+import { ECPair } from "../utils/ecpair";
+import { deriveKey } from "../utils/rescueFile";
+import { ChainSwap, SubmarineSwap } from "../utils/swapCreator";
+
+const mapSwap = (swap: RescuableSwap): SubmarineSwap | ChainSwap => {
+    if (swap.type === SwapType.Submarine) {
+        return {
+            ...swap,
+            type: SwapType.Submarine,
```
jackstar12, line 19: isnt this line redundant?

src/pages/RefundRescue.tsx
```diff
@@ -0,0 +1,87 @@
...
+            version: OutputType.Taproot,
+            address: swap.lockupAddress,
+            refundPrivateKeyIndex: swap.keyIndex,
+            claimPublicKey: swap.serverPublicKey,
+        } as Partial<SubmarineSwap> as SubmarineSwap;
+    }
+
+    return {
+        ...swap,
+        type: SwapType.Chain,
```
jackstar12, line 31: same as above, might wanna check it actually is of type `Chain` before aswell

jackstar12, line 31: same as above

e2e/bolt12.spec.ts
```diff
@@ -26,7 +27,7 @@ test.describe("BOLT12", () => {
         );
         await buttonCreateSwap.click();
 
-        const skipDownload = page.getByText("Skip download");
-        await expect(skipDownload).toBeVisible();
+        const downloadButton = page.getByText(dict.en.download_rescue_key);
```
kilrau, line 30: My bad

kilrau (approved):
Refunding in clean browser works nicely. 

image

Here the new rescue endpoint call:
image
image

maybeast (approved):

kilrau (approved):
Tested the flow e2e again on testnet - refunding works nicely on a fresh browser instance. I am glad we did the switch to the mnemonic, already got some ideas what we can do :tada:

PR #828, opened 2025-02-27
Refund: use 3rd party blockexplorer endpoints to double check if a lockup address has funds

### LINKED ISSUE #809: Refund: use 3rd party blockexplorer endpoints to double check if a lockup address has funds

### LINKED ISSUE #832: Rescue Key: Pagination, display swaps that were not funded as "No refund due"
Noticed today that the Rescue Key Refund flow in a fresh browser instance shows a swap I created (but not funded) with my Boltz Rescue Key to be available for refund but in a non-clickcable inactive state.

Image

Just like https://github.com/BoltzExchange/boltz-web-app/issues/809 we want Web App to double check lockup addresses of swaps by itself to also allow refunding of swaps whose lockup addresses were funded *after* the swap happened. Swap state can be success/failed/expired - everything.

Now the thing is that a rescue key can have thousands of swaps in its history and web app can't just hit mempool APIs with such a load without getting rate limited. So I propose to only check max e.g. 5 swaps back, if there are more swaps attached to the rescue key we need a "Load more" button to check the next 5. Something like that.

This something you could imagine working on? @kiocosta

kiocosta:
Closes #809
Closes #832

I have tested it this way:
1. Created the BTC->LN invoice
2. Made the HTLC expire by mining a bunch of blocks
3. Verified the swap is expired due to HTLC timeout -- not refundable yet
4. Sent sats to the lockup address
5. Verified the swap is now refundable
6. Added a refund address and refunded successfully

Did the above with L-BTC as well.

**Caveats**:
1. At this moment, only one UTXO is refundable through the web app. I will assign myself another issue to improve this.
2. It's also not possible to refund UTXOs of a successfully completed swap. Backend currently doesn't allow it: 
```
POST: http://localhost:9001/v2/swap/submarine/:swapId
Bad Request:
"{"error":"swap not eligible for a cooperative refund: status not eligible"}"
```

I believe these two edge-cases should be addressed on a different PR.

### DIFF
--- a/src/components/Pagination.tsx
+++ b/src/components/Pagination.tsx
@@ -1,0 +1,153 @@
+import { useSearchParams } from "@solidjs/router";
+import {
+    Accessor,
+    For,
+    Setter,
+    createMemo,
+    mergeProps,
+    onMount,
+} from "solid-js";
+
+import { useGlobalContext } from "../context/Global";
+import "../style/pagination.scss";
+
+const Pagination = <T,>(initialProps: {
+    items: Accessor<T[]>;
+    setDisplayedItems: (items: T[]) => void;
+    currentPage: Accessor<number>;
+    setCurrentPage: Setter<number>;
+    sort?: (items: T[]) => T[];
+    itemsPerPage?: number;
+    totalItems: number;
+}) => {
+    const props = mergeProps({ itemsPerPage: 15 }, initialProps);
+    const { t } = useGlobalContext();
+
+    const [searchParams, setSearchParams] = useSearchParams();
+
+    const updateDisplayedItems = () => {
+        const indexOfLastItem = props.currentPage() * props.itemsPerPage;
+        const indexOfFirstItem = indexOfLastItem - props.itemsPerPage;
+
+        let displayedItems = props.items();
+
+        if (typeof props.sort === "function") {
+            displayedItems = props.sort(displayedItems);
+        }
+
+        props.setDisplayedItems(
+            displayedItems.slice(indexOfFirstItem, indexOfLastItem),
+        );
+    };
+
+    onMount(() => updateDisplayedItems());
+
+    const numberOfPages = createMemo(() => {
+        const totalPages = Math.ceil(props.totalItems / props.itemsPerPage);
+        const numberOfPages = Array.from(
+            { length: totalPages },
+            (_, index) => index + 1,
+        );
+
+        let page = Number(searchParams.page);
+
+        if (isNaN(page) || page < 1 || page > numberOfPages.length) {
+            page = 1;
+            setSearchParams({ page });
+        }
+
+        props.setCurrentPage(page);
+
+        return numberOfPages;
+    });
+
+    const navigatePreviously = () => {
+        const page = props.setCurrentPage((prev: number) => prev - 1);
+        setSearchParams({ page });
+        updateDisplayedItems();
+    };
+
+    const navigateToNumber = (pageNumber: number) => {
+        const page = props.setCurrentPage(pageNumber);
+        setSearchParams({ page });
+        updateDisplayedItems();
+    };
+
+    const navigateNext = () => {
+        const page = props.setCurrentPage((prev: number) => prev + 1);
+        setSearchParams({ page });
+        updateDisplayedItems();
+    };
+
+    const pageNumbers = () => {
+        const current = props.currentPage();
+        const total = numberOfPages().length;
+
+        let start: number;
+        let end: number;
+
+        // Always displays a maximum of 5 page numbers
+        if (current <= 2) {
+            start = 0;
+            end = 5;
+        } else if (current >= total - 1) {
+            start = total - 5;
+            end = total;
+        } else {
+            start = current - 3;
+            end = current + 2;
+        }
+
+        return numberOfPages().slice(Math.max(0, start), Math.min(total, end));
+    };
+
+    return (
+        <>
+            <nav class="pagination">
+                <ul>
+                    <li>
+                        <button
+                            data-testid="previous-page"
+                            disabled={props.currentPage() === 1}
+                            onClick={navigatePreviously}>
+                            {t("back").toUpperCase()}
+                        </button>
+                    </li>
+                    <For each={pageNumbers()}>
+                        {(pageNumber) => (
+                            <li class="pagination-item">
+                                <button
+                                    data-testid={`page-${pageNumber}`}
+                                    class={`${props.currentPage() === pageNumber ? "active" : ""}`}
+                                    onClick={() =>
+                                        navigateToNumber(pageNumber)
+                                    }>
+                                    {pageNumber}
+                                </button>
+                            </li>
+                        )}
+                    </For>
+                    <li>
+                        <button
+                            data-testid="next-page"
+                            disabled={
+                                props.currentPage() ===
+                                numberOfPages()[numberOfPages().length - 1]
+                            }
+                            onClick={navigateNext}>
+                            {t("next").toUpperCase()}
+                        </button>
+                    </li>
+                </ul>
+            </nav>
+            <p>
+                {t("pagination_info", {
+                    start: props.currentPage,
+                    end: numberOfPages().length,
+                })}
+            </p>
+        </>
+    );
+};
+
+export default Pagination;
--- a/src/components/RefundButton.tsx
+++ b/src/components/RefundButton.tsx
@@ -1,5 +1,4 @@
 import { crypto } from "bitcoinjs-lib";
-import { OutputType } from "boltz-core";
 import { Signature, TransactionResponse } from "ethers";
 import { Network as LiquidNetwork } from "liquidjs-lib/src/networks";
 import log from "loglevel";
@@ -8,10 +7,10 @@ import {
     Setter,
     Show,
     createEffect,
+    createMemo,
     createResource,
     createSignal,
 } from "solid-js";
-import { ChainSwap, SubmarineSwap } from "src/utils/swapCreator";
 
 import RefundEta from "../components/RefundEta";
 import { RBTC } from "../consts/Assets";
@@ -19,15 +18,13 @@ import { SwapType } from "../consts/Enums";
 import { deriveKeyFn, useGlobalContext } from "../context/Global";
 import { usePayContext } from "../context/Pay";
 import { useWeb3Signer } from "../context/Web3";
-import {
-    getEipRefundSignature,
-    getLockupTransaction,
-} from "../utils/boltzClient";
+import { getEipRefundSignature } from "../utils/boltzClient";
 import { getAddress, getNetwork } from "../utils/compat";
 import { formatError } from "../utils/errors";
 import { decodeInvoice } from "../utils/invoice";
 import { refund } from "../utils/refund";
 import { prefix0x, satoshiToWei } from "../utils/rootstock";
+import { ChainSwap, SomeSwap, SubmarineSwap } from "../utils/swapCreator";
 import ContractTransaction from "./ContractTransaction";
 import LoadingSpinner from "./LoadingSpinner";
 
@@ -124,7 +121,7 @@ export const RefundBtc = (props: {
         t,
         deriveKey,
     } = useGlobalContext();
-    const { setSwap } = usePayContext();
+    const { setSwap, refundableUTXOs } = usePayContext();
 
     const [timeoutEta, setTimeoutEta] = createSignal<number | null>(null);
     const [timeoutBlockheight, setTimeoutBlockheight] = createSignal<
@@ -164,26 +161,30 @@ export const RefundBtc = (props: {
         setRefundRunning(true);
 
         try {
-            const res = await refund(
-                props.deriveKeyFn || deriveKey,
-                props.swap(),
-                refundAddress(),
-                lockupTransaction(),
-                true,
-                externalBroadcast(),
-            );
+            let refundedSwap: SomeSwap;
+
+            for (const tx of refundableUTXOs()) {
+                refundedSwap = await refund(
+                    props.deriveKeyFn || deriveKey,
+                    props.swap(),
+                    refundAddress(),
+                    tx,
+                    true,
+                    externalBroadcast(),
+                );
+            }
 
             // save refundTx into swaps json and set it to the current swap
             // only if the swap exist in localstorage, else it is a refund json
             // so we save it into the signal
-            const currentSwap = await getSwap(res.id);
+            const currentSwap = await getSwap(refundedSwap.id);
             if (currentSwap !== null) {
-                currentSwap.refundTx = res.refundTx;
+                currentSwap.refundTx = refundedSwap.refundTx;
                 await setSwapStorage(currentSwap);
                 setSwap(currentSwap);
             }
             if (props.setRefundTxId) {
-                props.setRefundTxId(res.refundTx);
+                props.setRefundTxId(refundedSwap.refundTx);
             }
 
             setRefundAddress("");
@@ -202,10 +203,13 @@ export const RefundBtc = (props: {
                     msg === "non-final"
                 ) {
                     msg = t("locktime_not_satisfied");
-                    setTimeoutEta(lockupTransaction().timeoutEta);
-                    setTimeoutBlockheight(
-                        lockupTransaction().timeoutBlockHeight,
+                    const legacyTx = refundableUTXOs().find(
+                        (tx) => tx.timeoutEta && tx.timeoutBlockHeight,
                     );
+                    if (legacyTx) {
+                        setTimeoutEta(legacyTx.timeoutEta);
+                        setTimeoutBlockheight(legacyTx.timeoutBlockHeight);
+                    }
                 }
                 log.error(msg);
                 notify("error", msg);
@@ -218,44 +222,18 @@ export const RefundBtc = (props: {
         setRefundRunning(false);
     };
 
-    // eslint-disable-next-line solid/reactivity
-    const [lockupTransaction] = createResource(props.swap, async (swap) => {
-        if (!swap) {
-            return undefined;
-        }
-
-        const transactionToRefund = await getLockupTransaction(
-            swap.id,
-            swap.type,
-        );
-
-        // show refund ETA for legacy swaps
-        if (swap.version !== OutputType.Taproot) {
-            setTimeoutEta(transactionToRefund.timeoutEta);
-            setTimeoutBlockheight(transactionToRefund.timeoutBlockHeight);
-        }
-
-        return transactionToRefund;
-    });
-
-    const buttonMessage = () => {
-        if (lockupTransaction.state == "errored") {
+    const buttonMessage = createMemo(() => {
+        if (refundableUTXOs()?.length === 0) {
             return t("no_lockup_transaction");
         }
         if (valid() || !refundAddress() || !props.swap()) {
             return t("refund");
         }
         return t("invalid_address", { asset: props.swap()?.assetSend });
-    };
+    });
 
     return (
-        <Show
-            when={
-                lockupTransaction.state === "ready" ||
-                lockupTransaction.state == "unresolved" ||
-                lockupTransaction.state == "errored"
-            }
-            fallback={<LoadingSpinner />}>
+        <Show when={refundableUTXOs()} fallback={<LoadingSpinner />}>
             <Show when={timeoutEta() > 0 || timeoutBlockheight() > 0}>
                 <RefundEta
                     timeoutEta={timeoutEta}
@@ -272,7 +250,7 @@ export const RefundBtc = (props: {
             <input
                 data-testid="refundAddress"
                 id="refundAddress"
-                disabled={lockupTransaction.state == "errored"}
+                disabled={refundableUTXOs().length === 0}
                 value={refundAddress()}
                 onInput={(e) => setRefundAddress(e.target.value.trim())}
                 type="text"
--- a/src/config.ts
+++ b/src/config.ts
@@ -31,6 +31,7 @@ const defaults = {
 
 type Asset = {
     blockExplorerUrl?: Url;
+    blockExplorerApis?: Url[];
 
     rifRelay?: string;
     contracts?: {
--- a/src/context/Pay.tsx
+++ b/src/context/Pay.tsx
@@ -7,6 +7,7 @@ import {
 } from "solid-js";
 import type { JSX } from "solid-js";
 
+import { LockupTransaction } from "../utils/boltzClient";
 import { SomeSwap } from "../utils/swapCreator";
 
 export type PayContextType = {
@@ -18,6 +19,12 @@ export type PayContextType = {
     setSwapStatus: Setter<string>;
     swapStatusTransaction: Accessor<SwapStatusTransaction>;
     setSwapStatusTransaction: Setter<SwapStatusTransaction>;
+    refundableUTXOs: Accessor<
+        (Partial<LockupTransaction> & Pick<LockupTransaction, "hex">)[]
+    >;
+    setRefundableUTXOs: Setter<
+        (Partial<LockupTransaction> & Pick<LockupTransaction, "hex">)[]
+    >;
 };
 
 const PayContext = createContext<PayContextType>();
@@ -36,6 +43,9 @@ const PayProvider = (props: { children: JSX.Element }) => {
     const [swapStatus, setSwapStatus] = createSignal<string>("");
     const [swapStatusTransaction, setSwapStatusTransaction] =
         createSignal<SwapStatusTransaction>({});
+    const [refundableUTXOs, setRefundableUTXOs] = createSignal<
+        (Partial<LockupTransaction> & Pick<LockupTransaction, "hex">)[]
+    >([]);
 
     return (
         <PayContext.Provider
@@ -48,6 +58,8 @@ const PayProvider = (props: { children: JSX.Element }) => {
                 setSwapStatus,
                 swapStatusTransaction,
                 setSwapStatusTransaction,
+                refundableUTXOs,
+                setRefundableUTXOs,
             }}>
             {props.children}
         </PayContext.Provider>
--- a/src/pages/Pay.tsx
+++ b/src/pages/Pay.tsx
@@ -4,6 +4,7 @@ import {
     Match,
     Show,
     Switch,
+    createEffect,
     createMemo,
     createResource,
     createSignal,
@@ -34,12 +35,15 @@ import TransactionClaimed from "../status/TransactionClaimed";
 import TransactionConfirmed from "../status/TransactionConfirmed";
 import TransactionLockupFailed from "../status/TransactionLockupFailed";
 import TransactionMempool from "../status/TransactionMempool";
-import { getSwapStatus } from "../utils/boltzClient";
+import { getLockupTransaction, getSwapStatus } from "../utils/boltzClient";
+import { getRefundableUTXOs, isSwapRefundable } from "../utils/refund";
+import { ChainSwap, SubmarineSwap } from "../utils/swapCreator";
 
 const Pay = () => {
     const params = useParams();
 
     const { getSwap, t } = useGlobalContext();
+
     const {
         swap,
         setSwap,
@@ -47,15 +51,20 @@ const Pay = () => {
         setSwapStatus,
         setSwapStatusTransaction,
         setFailureReason,
+        setRefundableUTXOs,
     } = usePayContext();
+
+    const prevSwapStatus = { value: "" };
+
+    createEffect(() => {
+        prevSwapStatus.value = swapStatus();
+    });
 
     createResource(async () => {
         const currentSwap = await getSwap(params.id);
 
         if (currentSwap) {
-            log.debug("selecting swap", currentSwap);
             setSwap(currentSwap);
-
             const res = await getSwapStatus(currentSwap.id);
             setSwapStatus(res.status);
             setSwapStatusTransaction(res.transaction);
@@ -63,10 +72,49 @@ const Pay = () => {
         }
     });
 
+    createResource(swapStatus, async () => {
+        const isInitialSwapState =
+            (swapStatus() === swapStatusPending.SwapCreated &&
+                prevSwapStatus.value === "") ||
+            (swapStatus() === swapStatusPending.InvoiceSet &&
+                prevSwapStatus.value === "");
+
+        // No need to fetch UTXO data for a swap just created
+        if (isInitialSwapState) {
+            return;
+        }
+
+        // We don't check the block explorer during the initial phase
+        // of a swap because, more often than not, it doesn't have
+        // information about the lockup transaction yet.
+        const shouldCheckBlockExplorer =
+            swapStatus() !== swapStatusPending.InvoiceSet &&
+            prevSwapStatus.value !== swapStatusPending.InvoiceSet &&
+            swapStatus() !== swapStatusPending.SwapCreated &&
+            prevSwapStatus.value !== swapStatusPending.SwapCreated &&
+            isSwapRefundable(swap());
+
+        try {
+            const utxos = shouldCheckBlockExplorer
+                ? await getRefundableUTXOs(swap() as ChainSwap | SubmarineSwap)
+                : [await getLockupTransaction(swap().id, swap().type)];
+
+            setRefundableUTXOs(utxos);
+
+            if (utxos.length > 0) {
+                // if there are remaining UTXOs, we consider we don't have a refundTx yet
+                setSwap({ ...swap(), refundTx: "" });
+            }
+        } catch (e) {
+            log.debug("error fetching UTXOs: ", e.stack);
+        }
+    });
+
     onCleanup(() => {
         log.debug("cleanup Pay");
         setSwap(null);
         setSwapStatus(null);
+        setRefundableUTXOs([]);
     });
 
     const [statusOverride, setStatusOverride] = createSignal<
--- a/src/pages/Refund.tsx
+++ b/src/pages/Refund.tsx
@@ -1,88 +1,96 @@
 import { useNavigate } from "@solidjs/router";
-import log from "loglevel";
-import { Show, createSignal, onMount } from "solid-js";
+import { Show, createResource, createSignal } from "solid-js";
 
-import SwapList from "../components/SwapList";
+import LoadingSpinner from "../components/LoadingSpinner";
+import Pagination from "../components/Pagination";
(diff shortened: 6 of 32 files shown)


kilrau (changes requested):
Doesn't work for me on Testnet. I created a Liquid -> Lightning Swap on Testnet, didn't send any funds and set the swap status manually to `invoice.failedToPay` in our backend. Then sent funds to the lockup address. WebApp still states "NO LOCKUP TRANSACTION FOUND". I don't think the calls to the explorers are being triggered.

[Screencast from 2025-03-04 11-44-42.webm](https://github.com/user-attachments/assets/54d88bd2-5bcd-40ca-a31f-7f0e0552c632)

This can be reproduced with a regtest setup the same way, just use `boltz-cli setswapstatus sWaPiD invoice.failedToPay` in the backend.

kilrau:
As discussed on our call, state of the swap doesn't matter, success/failed/expired. Users send funds to lockup addresses no matter the state of a swap. Probably can exclude `swap.created` as sending funds to a lockup address of  a swap in that state is simply continuing with the swap as intended.

kilrau:
As discussed on our call, state of the swap doesn't matter, success/failed/expired. Users send funds to lockup addresses no matter the state of a swap. Probably can exclude swap.created as sending funds to a lockup address of a swap in that state is simply continuing with the swap.

kiocosta:
@kilrau I have updated the description of the PR with the changes made.
I have also re-tested it with the new changes and it should now work with `invoice.failedToPay`.

kilrau:
Not sure what you changed since yesterday as yesterday web app called blockstream.info/liquid/api and detected the UTXO, but Just ran latest commit `4203bee` on mainnet with the exact same Liquid submarine swap (state `invoice.failedToPay`) that has a confirmed UTXO but now it doesn't call blockstream.info and thus doesn't detect the utxo: 
Screenshot from 2025-03-07 11-19-24

kilrau (changes requested):
Found the issue: calling blockstream.info works when refunding from browser storage, it does NOT work when using external refund file. That worked yesterday. Please test both scenarios. Also please test all other states apart from `invoice.failedToPay`.

kilrau:
Broadcasting the refund tx on liquid testnet works, on liquid mainnet not (min relay fee not met):
image

That sth to do with the switch to DiscountCT? If so, how can Caio make it work? @michael1011

kilrau:
Also Caio and I discussed that eventually we want web app to be able to handle refunds for submarine/chain swaps in *all* states, especially successful ones (`transaction.claimed`) to catch the case where users send to an old lockup address. Right now backend refuses to do this cooperatively `"{"error":"swap not eligible for a cooperative refund: status not eligible"}"`.

We can leave this for another PR, but is there anything speaking against backend re-checking a lockup address if it gets a cooperative refund request and allow cooperative refunds if there are funds in there? @michael1011

michael1011:
> That sth to do with the switch to DiscountCT?

No

You are trying to refund an input that the backend does not have, it thinks that it is not swap related. Therefore, it does not allow lowball. Never has. Discount CT fixes this

michael1011:
> but is there anything speaking against backend re-checking a lockup address if it gets a cooperative refund request and allow cooperative refunds if there are funds in there?

The logic to decide which things to sign cooperatively gets ***a lot more*** complicated. Let's not in this PR

kilrau:
> > That sth to do with the switch to DiscountCT?
> 
> No
> 
> You are trying to refund an input that the backend does not have, it thinks that it is not swap related. Therefore, it does not allow lowball. Never has. Discount CT fixes this

Ahhh got it. So once https://github.com/BoltzExchange/boltz-web-app/pull/825 is merged a simple rebase should do the trick.

kilrau:
> > but is there anything speaking against backend re-checking a lockup address if it gets a cooperative refund request and allow cooperative refunds if there are funds in there?
> 
> The logic to decide which things to sign cooperatively gets _**a lot more**_ complicated. Let's not in this PR

Agreed. Opened as follow up issues: https://github.com/BoltzExchange/boltz-web-app/issues/834 & https://github.com/BoltzExchange/boltz-backend/issues/839

What we should do in this PR though is to try non-cooperative refund. Like this at least it'll work even though users might have to wait for the swap to expire. And we need that fallback anyways.

kiocosta:
@kilrau @michael1011 
Pushed new changes. Tested refund in the following scenarios:
- using rescue key
- using old refund file
- expired HTLC
- expired invoice
- invoice paid with incorrect amount
- invoice setted to "failedToPay" through the CLI

I was able to refund the UTXO in all of the above.

From `swapStatusFailed`, the only case I don't know how to test is `transaction.failed`.
UTXOs in `transaction.refunded` or `swap.refunded` are still not refundable, as written in the beginning of the PR.

I ran all e2e tests locally. Sometimes they would break when running in parallel, but all passed when ran in isolation (not sure if this is something to be concerned or not).

kilrau:
> From `swapStatusFailed`, the only case I don't know how to test is `transaction.failed`.

CTRL+F in https://docs.boltz.exchange/api/lifecycle then you know when this state happens. For testing you can just set it manually on regtest backend via `boltz-cli setswapstatus sWaPiD transaction.failed`.

Once you figured out the e2e thing and CI is passing I'll do a final round of manual testing. @michael1011 an idea?

src/utils/refund.ts
```diff
@@ -38,7 +38,7 @@ const refundTaproot = async <T extends TransactionInterface>(
     privateKey: ECPairInterface,
     decodedAddress: DecodedAddress,
     feePerVbyte: number,
-    timeoutBlockHeight: number,
+    timeoutBlockHeight?: number,
```
michael1011, line 41: We *need* that information for refunds that are not cooperative

src/utils/swapCreator.ts
```diff
@@ -20,6 +20,7 @@ import { getPair } from "./helper";
 export type SwapBase = {
     type: SwapType;
     status?: string;
+    address?: string;
```
michael1011, line 23: Why add `address` here?

src/components/RefundButton.tsx
```diff
@@ -223,19 +227,38 @@ export const RefundBtc = (props: {
...
+            const utxos = await fetchUTXOsWithFailover(
+                swap.assetSend,
+                swap.address,
+            );
 
-        return transactionToRefund;
+            if (utxos.length) {
+                const rawTx = await fetchRawTxWithFailover(
+                    swap.assetSend,
+                    utxos[0].txid,
```
michael1011, line 252: Why only take the first one?

src/components/RefundButton.tsx
```diff
@@ -223,19 +227,38 @@ export const RefundBtc = (props: {
...
-        }
+            return transactionToRefund;
+        } catch {
+            const utxos = await fetchUTXOsWithFailover(
+                swap.assetSend,
+                swap.address,
+            );
 
-        return transactionToRefund;
+            if (utxos.length) {
```
michael1011, line 249:
```suggestion
            if (utxos.length > 0) {
```

Not a fan of relying on 0 being falsy

src/pages/RefundExternal.tsx
```diff
@@ -111,8 +112,26 @@ export const RefundBtcLike = () => {
...
+                res.map(async (swap) => {
+                    const utxos = await fetchUTXOsWithFailover(
+                        swap.symbol,
+                        swap.lockupAddress,
+                    );
+                    if (utxos.length) {
+                        return {
+                            ...swap,
+                            transaction: {
+                                id: utxos[0].txid,
```
michael1011, line 125: Why only take the first one?

src/pages/RefundExternal.tsx
```diff
@@ -111,8 +112,26 @@ export const RefundBtcLike = () => {
...
+                                vout: null,
```
michael1011, line 126: The response gives us the `vout`; let's use it

src/pages/RefundExternal.tsx
```diff
@@ -111,8 +112,26 @@ export const RefundBtcLike = () => {
             }
 
             const res = await getRescuableSwaps(getXpub(source.refundJson));
-            rescueContext.setRescuableSwaps(res);
-            return res;
+            const swapsWithUTXO = await Promise.all(
+                res.map(async (swap) => {
```
michael1011, line 116: Doing this for ***all** swaps in your history, will get you rate limit and/or your IP address banned from the block explorer. @kilrau I guess we should paginate this.

src/pages/Refund.tsx
```diff
@@ -29,19 +30,32 @@ const Refund = () => {
...
 
-        const swapsToRefund = allSwaps
-            .filter(refundSwapsSanityFilter)
-            .filter(
-                (swap) =>
-                    swapStatusFailed.TransactionLockupFailed === swap.status,
-            );
+        const swapsWithUTXO = (
+            await Promise.all(
+                allSwaps.filter(refundSwapsSanityFilter).map(async (swap) => {
```
michael1011, line 35: Same thing here. You cannot hit the block explorer for every swap. It will get you rate limited

src/status/SwapExpired.tsx
```diff
@@ -22,6 +26,19 @@ const SwapExpired = () => {
...
         } catch (e) {
+            const utxos = await fetchUTXOsWithFailover(
+                swap().assetSend,
+                swap().address,
+            );
+
+            if (utxos.length) {
+                const rawTx = await fetchRawTxWithFailover(
+                    swap().assetSend,
+                    utxos[0].txid,
```
michael1011, line 37: Why only the first one?

michael1011 (changes requested):
> Only InvoiceFailedToPay and InvoicePending seem to be allowed for this command.

Yes. Changing the state to `transaction.failed` could lead to loss of funds in prod, which is why the command doesn't let you do it. Either change the status in the database and restart the backend or have less funds in the wallet the backend uses than it attempts to send

src/status/SwapExpired.tsx
kiocosta: @michael1011 currently we consider `refundTx` as a string
```
export type SwapBase = {
    ...
    refundTx?: string;
    ...
```
and use it as a condition to decide on what to render in different components: `RefundEvm`, `RefundExternal`, `SwapRefunded` and `Pay`. 

I think changing the logic of these to deal with multiple `refundTxs` may lead to more complexity in the PR, so I thought I'd do it in another PR as a way of keeping things simpler for reviewing and testing.

However, I can implement it on this PR if preferred.

src/config.ts
```diff
@@ -31,6 +31,7 @@ const defaults = {
 
 type Asset = {
     blockExplorerUrl?: Url;
+    apiUrls?: Url[];
```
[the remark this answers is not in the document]
kiocosta: @michael1011 done

src/utils/blockchain.ts
```diff
@@ -0,0 +1,77 @@
...
+    }[];
+    size: number;
+    weight: number;
+    fee: number;
+    status: {
+        confirmed: boolean;
+        block_height: number;
+        block_hash: string;
+        block_time: number;
+    };
```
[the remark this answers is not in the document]
kiocosta: @michael1011 fixed.

src/utils/blockchain.ts
```diff
@@ -0,0 +1,77 @@
+import { chooseUrl, config } from "../config";
+
+export interface Transaction {
```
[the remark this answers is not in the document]
kiocosta: @michael1011 fixed.

src/utils/blockchain.ts
```diff
@@ -0,0 +1,77 @@
...
+    address: string,
+): Promise<Transaction[]> => {
+    for (const url of config.assets[asset].apiUrls) {
+        try {
+            const basePath = chooseUrl(url);
+            const response = await fetch(`${basePath}/address/${address}/utxo`);
+
+            if (!response.ok) {
+                continue;
+            }
```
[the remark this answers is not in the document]
kiocosta: @michael1011 logged with:
```
            if (!response.ok) {
                log.error(`Failed to fetch UTXOs for ${address}`);
                continue;
            }

            return (await response.json()) as UTXO[];
        } catch (e) {
            log.error(`Failed to fetch UTXOs for ${address}: ${e.stack}`);
            continue;
        }
```
I noticed `e.stack` is not commonly used on the project. I like being able to see the callstack, what do you think?
image

src/utils/blockchain.ts
```diff
@@ -0,0 +1,77 @@
...
+    txid: string,
+): Promise<{ hex: string }> => {
+    for (const url of config.assets[asset].apiUrls) {
+        try {
+            const basePath = chooseUrl(url);
+            const response = await fetch(`${basePath}/tx/${txid}/hex`);
+
+            if (!response.ok) {
+                continue;
+            }
```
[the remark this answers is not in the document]
kiocosta: @michael1011 done like the one above.

src/utils/blockchain.ts
```diff
@@ -0,0 +1,77 @@
...
+
+            return (await response.json()) as Transaction[];
+        } catch {
+            continue;
+        }
+    }
+
+    throw new Error("All API endpoints failed");
```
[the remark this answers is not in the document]
kiocosta: @michael1011 updated

src/utils/boltzClient.ts
```diff
@@ -126,6 +126,7 @@ type SubmarineCreatedResponse = {
 type ReverseCreatedResponse = {
     id: string;
     invoice: string;
+    address: string;
```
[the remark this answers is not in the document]
kiocosta: @michael1011 indeed, removed.

src/utils/refund.ts
kiocosta: @michael1011 fixed

src/utils/swapCreator.ts
kiocosta: @michael1011 I thought it was missing because I use it in a few places to get the UTXOs:
```
const utxos = await fetchUTXOsWithFailover(
    swap.assetSend,
    swap.address,
);
```
But this is actually only applicable to `SubmarineSwap`s, right?
I have added the typing of `SubmarineSwap` in the refund workflows used in this PR.

src/components/RefundButton.tsx
kiocosta: @michael1011 fixed this and the other cases where this is repeated

src/pages/RefundExternal.tsx
kiocosta: @michael1011 vout added.

src/pages/RefundExternal.tsx
kiocosta: @michael1011 @kilrau working on the pagination right now.

kiocosta:
@michael1011 @kilrau 

I've pushed new changes that add pagination to the refund page and address all of Michael's comments on the PR, except for those related to refunding multiple UTXOs.

I have WIP for refunding multiple UTXOs. I need more time to complete and test it, as it involves changes across multiple components.

Current pagination looks and behavior:

https://github.com/user-attachments/assets/8b16efc2-ffb2-4334-a7c3-4ef7239df7cb

@kilrau There are new translations that need review.

I chose the label "No refund due" instead of "Non-refundable" to avoid confusion. "Non-refundable" might imply that a swap can never be refunded, whereas "No refund due" clarifies that there's nothing to refund at the moment.

kilrau:
Looking good! :muscle: Will do more in-depth testing and push strings tomorrow.

kilrau (changes requested):
Tried all sort of scenarios (creating many swaps, close browser then send to a swap address of one of them) and web app reliably and immediately detected the swap as refundable when I reopened the page. Good!

One important edge case that should be fixed in this PR: If I refund a swap and then send funds AGAIN to the lockup address of this swap, web app correctly detects the UTXO and marks the swap as refundable BUT it doesn't reset the swap pages state to let me refund *again*:
[Screencast from 2025-03-14 12-31-10.webm](https://github.com/user-attachments/assets/8c2bf1bb-409f-49fe-9d88-0fb3a4dbc859)

Also here a list of some UI/UX issues:
1. `Refund`/`No Refund Due` Button disappears on mobile, prbly not enough space. Or did you intend that? I'd rather get rid of the date to make it fit and keep the button, the button is too important as it serves two purposes now: signify if the swap is refundable or not and well signify that one can tab on it to refund. Gotta think of sth. image
2. Sticky pagination menu overlay when opening hamburger menu on mobile:
[Screencast from 2025-03-14 12-15-11.webm](https://github.com/user-attachments/assets/44a0d4d2-86cc-43b9-a06f-af1b3a74f577)
3. Pagination should be added as `/1`, `/2` etc slug to the url, so I can refresh in any state.

Just to say it again: I opened a [separate issue for refunding multiple UTXOs](https://github.com/BoltzExchange/boltz-web-app/issues/851), so let's take this out and focus on fixing the above and getting this PR merged.

Strings fixed. One more round :muscle:

kilrau:
Also pls check the failing e2e tests. Maybe try a rebase.

src/components/Pagination.tsx
```diff
@@ -0,0 +1,132 @@
...
+    currentPage: Accessor<number>;
+    setCurrentPage: Setter<number>;
+    sort?: (items: unknown[]) => unknown[];
+    itemsPerPage: number;
+    totalItems: number;
+}) => {
+    const { t } = useGlobalContext();
+    const [numberOfPages, setNumberOfPages] = createSignal<number[]>([]);
+
+    onMount(() => {
```
michael1011, line 19: This shouldn't be done `onMount` but with `createMemo`

src/components/Pagination.tsx
```diff
@@ -0,0 +1,132 @@
...
+        const numberOfPages = [];
+        for (
+            let i = 1;
+            i <= Math.ceil(props.totalItems / props.itemsPerPage);
+            i++
+        ) {
+            numberOfPages.push(i);
+        }
```
michael1011, line 27:
```suggestion
const totalPages = Math.ceil(props.totalItems / props.itemsPerPage);
const numberOfPages = Array.from({ length: totalPages }, (_, index) => index + 1);
```

src/components/Pagination.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+import { useGlobalContext } from "../context/Global";
+import "../style/pagination.scss";
+
+const Pagination = (props: {
+    items: Accessor<unknown[]>;
+    setDisplayedItems: (items: unknown[]) => void;
+    currentPage: Accessor<number>;
+    setCurrentPage: Setter<number>;
+    sort?: (items: unknown[]) => unknown[];
```
michael1011, line 12: Use a generic instead of `unknown` to get type safety

src/pages/RefundExternal.tsx
michael1011: This one needs pagination too

src/pages/Refund.tsx
```diff
@@ -1,88 +1,145 @@
...
+import { swapStatusFailed, swapStatusPending } from "../consts/SwapStatus";
 import { useGlobalContext } from "../context/Global";
 import "../style/tabs.scss";
+import { fetchUTXOsWithFailover } from "../utils/blockchain";
 import { getLockupTransaction, getSwapStatus } from "../utils/boltzClient";
-import { SomeSwap } from "../utils/swapCreator";
+import { SomeSwap, SubmarineSwap } from "../utils/swapCreator";
 import ErrorWasm from "./ErrorWasm";
 
+const SWAPS_PER_PAGE = 5;
```
michael1011, line 18:
```suggestion
const swapsPerPage = 15;
```

michael1011 (commented):
Starting to look good!

src/components/Pagination.tsx
kiocosta: @michael1011 done

src/components/Pagination.tsx
kiocosta: @michael1011 done

src/pages/Refund.tsx
kiocosta: @michael1011 done

src/components/Pagination.tsx
kiocosta: @michael1011 memoized `numberOfPages` and used `onMount` just to invoke `updateDisplayedItems` on the first render.

kiocosta:
@kilrau 
>One important edge case that should be fixed in this PR: If I refund a swap and then send funds AGAIN to the lockup address of this swap, web app correctly detects the UTXO and marks the swap as refundable BUT it doesn't reset the swap pages state to let me refund again:

Fixed! Now the refund button is always available when there is any UTXO. When the user clicks the refund button, it also refunds **all** available UTXOs in that swap at that moment. I think all we have to do in #851 now is to display all the refund transaction ids, since right now I'm only displaying the last refund transaction id, but the refund of multiple UTXOs is already happening.

Fixed all the UX issues you listed as well.
For **4**, instead of adding `/1`, `/2`, I decided to use `?page=1`, `?page=2` etc, since AFAIK using the slug would require me to use [SolidJS's navigate](https://docs.solidjs.com/solid-router/reference/primitives/use-navigate) to update it. Although `navigate` updates super quick, it makes the page scroll to the top, which can be annoying to users. With query parameters I can update the list and the parameter without causing a scroll to the top. 

An invalid `?page=` value redirects user to `?page=1`.

src/pages/Refund.tsx
```diff
@@ -1,88 +1,152 @@
...
 import { useGlobalContext } from "../context/Global";
 import "../style/tabs.scss";
+import { fetchUTXOsWithFailover } from "../utils/blockchain";
 import { getLockupTransaction, getSwapStatus } from "../utils/boltzClient";
-import { SomeSwap } from "../utils/swapCreator";
+import { isMobile } from "../utils/helper";
+import { SomeSwap, SubmarineSwap } from "../utils/swapCreator";
 import ErrorWasm from "./ErrorWasm";
 
+const SWAPS_PER_PAGE = 10;
```
michael1011, line 19:
```suggestion
const swapsPerPage = 10;
```

src/utils/blockchain.ts
```diff
@@ -0,0 +1,80 @@
...
+        hex: string;
+        timeoutEta?: number;
+        timeoutBlockHeight: number;
+    }[] = [];
+
+    for (const utxo of utxos) {
+        const rawTx = await fetchRawTxWithFailover(swap.assetSend, utxo.txid);
+        txs.push({
+            hex: rawTx.hex,
+            timeoutBlockHeight: null,
```
michael1011, line 75:
```suggestion
            timeoutBlockHeight: swap.timeoutBlockHeight,
```

src/utils/blockchain.ts
```diff
@@ -0,0 +1,80 @@
...
+};
+
+export const getSwapUTXOs = async (swap: SubmarineSwap) => {
+    const utxos = await fetchUTXOsWithFailover(swap.assetSend, swap.address);
+
+    const txs: {
+        hex: string;
+        timeoutEta?: number;
+        timeoutBlockHeight: number;
+    }[] = [];
```
michael1011, line 69: Use the `LockupTransaction` from `boltzClient`

src/pages/RefundRescue.tsx
```diff
@@ -49,25 +58,67 @@ const mapSwap = (
...
     const [refundTxId, setRefundTxId] = createSignal<string>("");
 
+    createResource(async () => {
+        if (rescuableSwap()) {
+            const res = await getSwapStatus(rescuableSwap().id);
+            setSwapStatus(res.status);
+            setSwapStatusTransaction(res.transaction);
+            setFailureReason(res.failureReason);
+
+            if (rescuableSwap().type === SwapType.Submarine) {
```
michael1011, line 85: Why only for submarine swaps? Should also be done for chain swaps
Also, this does not consider swaps from RBTC

src/context/Pay.tsx
```diff
@@ -36,6 +44,9 @@ const PayProvider = (props: { children: JSX.Element }) => {
     const [swapStatus, setSwapStatus] = createSignal<string>("");
     const [swapStatusTransaction, setSwapStatusTransaction] =
         createSignal<SwapStatusTransaction>({});
+    const [refundableUTXOs, setRefundableUTXOs] = createSignal<Transaction[]>(
+        [],
+    );
```
michael1011, line 49: Why is this in the context? The flow is really unclear which side effects fetches which values when. And the initial value for this is `[]` but you set it `null` in some places.

src/pages/Pay.tsx
```diff
@@ -34,39 +35,64 @@ import TransactionClaimed from "../status/TransactionClaimed";
...
         if (currentSwap) {
-            log.debug("selecting swap", currentSwap);
-            setSwap(currentSwap);
-
             const res = await getSwapStatus(currentSwap.id);
             setSwapStatus(res.status);
             setSwapStatusTransaction(res.transaction);
             setFailureReason(res.failureReason);
+
+            if (currentSwap.type === SwapType.Submarine) {
```
michael1011, line 67: This is effectively duplicate code to `src/pages/RefundRescue.tsx`

src/pages/Pay.tsx
```diff
@@ -34,39 +35,64 @@ import TransactionClaimed from "../status/TransactionClaimed";
...
+                try {
+                    const utxos = await getSwapUTXOs(
+                        currentSwap as SubmarineSwap,
+                    );
+                    setRefundableUTXOs(utxos);
```
michael1011, line 72: Waiting on a block explorer API to respond to show a swap is a ***bad*** idea

src/pages/Refund.tsx
```diff
@@ -1,88 +1,152 @@
...
+
+                        if (
+                            !(await updateSwapStatus(swap.id, res.status)) &&
+                            Object.values(swapStatusFailed).includes(res.status)
+                        ) {
+                            // Make sure coins were locked for the swaps with status "swap.expired" or "swap.failedToPay"
+                            const lockupTransaction =
+                                await getLockupTransaction(swap.id, swap.type);
+
+                            const utxos = await fetchUTXOsWithFailover(
```
michael1011, line 60: We ***cannot*** rely on the block explorer APIs to work

src/pages/RefundExternal.tsx
```diff
@@ -47,20 +58,46 @@ enum RefundType {
...
+            log.debug("selecting swap", swap);
+            return swap;
+        }
+        return null;
+    });
+
     return (
         <>
-            <Show when={refundTxId() === ""}>
+            <Show when={refundTxId() === "" && refundableSwap() !== undefined}>
```
michael1011, line 97: You return null but check for undefined

src/pages/RefundExternal.tsx
```diff
@@ -47,20 +58,46 @@ enum RefundType {
...
     const { t } = useGlobalContext();
+    const { setRefundableUTXOs } = usePayContext();
 
     const [refundTxId, setRefundTxId] = createSignal<string>("");
 
+    const swap = createMemo(() => props.refundJson());
+
+    const [refundableSwap] = createResource(swap, async (swap) => {
+        if (swap) {
+            if (swap.type === SwapType.Submarine) {
```
michael1011, line 76: Same thing here. This also applies to chain swaps and should exclude swaps from RBTC

src/pages/RefundExternal.tsx
```diff
@@ -112,7 +149,58 @@ export const RefundBtcLike = () => {
...
+                            Object.values(swapStatusFailed).includes(res.status)
+                        ) {
+                            // Make sure coins were locked for the swaps with status "swap.expired" or "swap.failedToPay"
+                            const lockupTransaction =
+                                await getLockupTransaction(swap.id, swap.type);
+
+                            const utxos = await fetchUTXOsWithFailover(
+                                swap.assetSend,
+                                swap.address,
+                            );
```
michael1011, line 175: This needs pagination. Fetching transactions UTXOs for all swaps will get you banned

src/pages/RefundExternal.tsx
```diff
@@ -112,7 +149,58 @@ export const RefundBtcLike = () => {
...
+
+                            if (lockupTransaction && utxos.length > 0) {
+                                return swap;
+                            }
+
+                            if (utxos.length > 0) {
+                                return swap;
+                            }
```
michael1011, line 183: We ***cannot*** rely on the block explorer API working

michael1011 (changes requested):

kilrau:
> One important edge case that should be fixed in this PR: If I refund a swap and then send funds AGAIN to the lockup address of this swap, web app correctly detects the UTXO and marks the swap as refundable BUT it doesn't reset the swap pages state to let me refund _again_

This works now, awesome!

> 1. `Refund`/`No Refund Due` Button disappears on mobile, prbly not enough space.

That too, looking good (even on an iPhone SE).

> 2\. Check the spacing for different languages, looks like it's proportinal to string length and just doesn't look right for German, Spanish et al

Looking good too.

> 3\. Sticky pagination menu overlay when opening hamburger menu on mobile

Fixed.

> 4\. Pagination should be added
> Instead of adding /1, /2, I decided to use ?page=1, ?page=2 etc... AFAIK using the slug would require me to use [SolidJS's navigate](https://docs.solidjs.com/solid-router/reference/primitives/use-navigate) to update it. Although navigate updates super quick, it makes the page scroll to the top, which can be annoying to users IMO. With query parameters I can update the list and the parameter without causing a scroll to the top.

Works!

> When the user clicks the refund button, it also refunds ALL available UTXOs in that swap at that moment.
> All we could do in https://github.com/BoltzExchange/boltz-web-app/issues/851 now is to display all the refund transaction ids, since right now I'm only displaying the last one, but the refund of multiple UTXOs is already happening!

As @michael1011 not in this PR to not add any fixes on topo, but this needs an urgent fix *after* merge: https://github.com/BoltzExchange/boltz-web-app/issues/855 let's tak

> Added the pagination to the `RefundExternal` page. Re-tested the refund flows multiple times in different scenarios.

Works nicely. Also tested reusing and old swap address there and was detected reliably.

src/context/Pay.tsx
kiocosta: @michael1011 I believe this should be in a context because I need the information about the UTXOs in many components, but I don't wanna prop drill it or perform duplicated requests for the block explorers.

We have three "refund flows": `Pay.tsx`, `RefundExternal.tsx` or `RefundRescue.tsx`. 
Each of these is a page which has the `RefundButton` component.

In the case of `Pay.tsx`, the `RefundButton` appear as a child of different components (`InvoiceFailedToPay`, `TransactionLockupFailed`, etc).

I need the information of the UTXOs at the page level, because I use it to conditionally render the `RefundButton`:
https://github.com/BoltzExchange/boltz-web-app/blob/2060d80ef1f8e012affa70132fdc5f5db5a77f1a/src/pages/Pay.tsx#L73-L76
https://github.com/BoltzExchange/boltz-web-app/blob/2060d80ef1f8e012affa70132fdc5f5db5a77f1a/src/pages/Pay.tsx#L114-L126

But I also need it on `RefundButton`, to perform the `refundAction` and to properly render the component.

Since the UTXO information is on the context, I don't have to make another GET request on `RefundButton` to use it. I also don't need to "prop drill" it: `Pay.tsx` -> `TransactionLockupFailed` -> `RefundButton`.

What do you think?

src/pages/RefundExternal.tsx
kiocosta: @michael1011 it's already using pagination:
https://github.com/BoltzExchange/boltz-web-app/blob/2060d80ef1f8e012affa70132fdc5f5db5a77f1a/src/pages/RefundExternal.tsx#L310-L320

This `createResource` uses the `currentSwaps` set by `Pagination` and only performs requests for these:
https://github.com/BoltzExchange/boltz-web-app/blob/2060d80ef1f8e012affa70132fdc5f5db5a77f1a/src/pages/RefundExternal.tsx#L156-L160

src/context/Pay.tsx
michael1011: Organizing this nicer would be great. Also we can surely avoid the code duplication to avoid having to maintain this code in three places

src/pages/Refund.tsx
kiocosta: @michael1011 fixed!

src/utils/blockchain.ts
kiocosta: @michael1011 this file actually had bad typing from my side. `fetchRawTxWithFailover` does not return `timeoutBlockHeight`. I have fixed the types for it tho (more on comment below).

src/utils/blockchain.ts
kiocosta: @michael1011 the return of `fetchRawTxWithFailover` doesn't match `LockupTransaction`, `hex` is the only property in common. I have typed them this way though:
https://github.com/BoltzExchange/boltz-web-app/blob/57f86b172334bb5ff518afbb78d15828716ddd29/src/utils/blockchain.ts#L87-L90

https://github.com/BoltzExchange/boltz-web-app/blob/57f86b172334bb5ff518afbb78d15828716ddd29/src/context/Pay.tsx#L22-L24

The intention is to keep them "tied together" and accomodate the original `lockupTransaction` and `utxos` on the same array.

src/pages/RefundRescue.tsx
kiocosta: @michael1011 fixed

src/pages/RefundExternal.tsx
kiocosta: @michael1011 fixed

kilrau:
> * no longer relying on 3rd party APIs to work. If they're up we use them, if not we maintain the existing behavior.

I tested this again with all scenarios I could think of with blockstream.info and liquid.network blocked:
image

Works :fire:

kilrau (approved):
> * no longer relying on 3rd party APIs to work. If they're up we use them, if not we maintain the existing behavior.

I tested this again with all scenarios I could think of with blockstream.info and liquid.network blocked. Works :fire:

kilrau:
Used this PR's Vercel deployment in a support ticket recently to rescue several lockup txes that web app didn't detect. Worked all as expected.

PR #836, opened 2025-03-08
feat: adjust network fee when RIF is needed

### LINKED ISSUE #673: Adjust fee for RIF Relay when connecting empty wallet
Adjust Network fee to include RIF Relay's fee if RBTC in connected wallet too low for claim. And it should throw a warning similiar to "Network fee changed".

michael1011:
Closes #673

Translations need a sanity check


maybeast (approved):

kilrau (approved):
Tested on mainnet with wallets containing enough and no RBTC - works as expected. Minor string improvements pushed.

Issue #859, opened 2025-03-20
Lightning address handling assumes no other query parameters


djkazic:
Here is code that properly handles the query params. If there are other query params the `?` is inappropriate.

Fixed code:

```
const fetchLnurlInvoice = async (amount: number, data: LnurlResponse) => {
    const separator = data.callback.includes('?') ? '&' : '?';
    const url = `${data.callback}${separator}amount=${amount}`;

    log.debug("fetching invoice", url);
    
    const res = await fetch(url).then(checkResponse<LnurlCallbackResponse>);
    
    log.debug("fetched invoice", res);
    
    return res.pr;
};
```

kilrau:
Nice! How about a PR instead?

djkazic:
Sure. Will open one now.

PR #860, opened 2025-03-20
fix: conditionally set amount query param for LN address fetch

djkazic:
Closes https://github.com/BoltzExchange/boltz-web-app/issues/859

### DIFF
--- a/src/utils/invoice.ts
+++ b/src/utils/invoice.ts
@@ -178,9 +178,14 @@ const checkLnurlResponse = (amount: number, data: LnurlResponse) => {
     return data;
 };
 
-const fetchLnurlInvoice = async (amount: number, data: LnurlResponse) => {
-    log.debug("fetching invoice", `${data.callback}?amount=${amount}`);
-    const res = await fetch(`${data.callback}?amount=${amount}`).then(
+export const fetchLnurlInvoice = async (
+    amount: number,
+    data: LnurlResponse,
+) => {
+    const url = new URL(data.callback);
+    url.searchParams.set("amount", amount.toString());
+    log.debug("fetching invoice", url.toString());
+    const res = await fetch(url.toString()).then(
         checkResponse<LnurlCallbackResponse>,
     );
     log.debug("fetched invoice", res);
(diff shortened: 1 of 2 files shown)


michael1011 (changes requested):
Thank you for the PR!

Let's use (`URL`)[https://developer.mozilla.org/en-US/docs/Web/API/URL] and modify its `searchParams` instead of messing with strings. Also, I test would be nice to avoid regressions

djkazic:
I've implemented the requested changes, PTAL.

michael1011 (changes requested):
LGTM

Please can `npm run prettier:write` for formatting, can be merged afterwards

kilrau (changes requested):
Thank you indeed! Test needs a smol fix, then should be good to go.

```
Error: tests/utils/fetchInvoice.spec.ts(41,60): error TS2345: Argument of type '{ callback: string; }' is not assignable to parameter of type 'LnurlResponse'.
  Type '{ callback: string; }' is missing the following properties from type 'LnurlResponse': minSendable, maxSendable
Error: Process completed with exit code 2.
```

djkazic:
Ah, okay. I will fix the formatting and the tests momentarily.

djkazic:
Force pushed a fix to the test.

djkazic:
Is the fix for `sendAmount` in scope for this PR? It seems like the culprit may be the handling of when `destinationAsset` is LN, we ignore the `sendAmount` query param. Instead, we rely on trying to parse the invoice amount which will fail since the lightning address is not an actual invoice.

My immediate thought is that we could fix that by detecting an LN address and fetching an invoice + calling `setInvoice` on that. Let me know what you think.

michael1011 (approved):

michael1011:
> Is the fix for sendAmount in scope for this PR? 

Let's do it in a separate one

Thank you for the fix!

kilrau:
> My immediate thought is that we could fix that by detecting an LN address and treating that differently -- in other words, making the comparison `if (destinationAsset !== LN && notLNAddress)`. That way, we could continue to parse `sendAmount` from the query param and set it in `Create.tsx`.

Something like that yes

PR #863, opened 2025-03-24
fix: parse query params when destination is an LN address

djkazic:
This PR adds a function `isLightningAddress` to more accurately determine if we should handle a given `destination` as eligible for parsing the `sendAmount` query parameter.

Previously the logic dictated that any `LN` type destination should defer to the invoice's embedded amount -- but this did not work for LN addresses.

### DIFF
--- a/src/context/Create.tsx
+++ b/src/context/Create.tsx
@@ -16,13 +16,14 @@ import { BTC, LBTC, LN, RBTC, assets } from "../consts/Assets";
 import { Side, SwapType } from "../consts/Enums";
 import { DictKey } from "../i18n/i18n";
 import { getAddress, getNetwork } from "../utils/compat";
+import { isLnurl } from "../utils/invoice";
 import { getUrlParam, urlParamIsSet } from "../utils/urlParams";
 
 const setDestination = (
     setAssetReceive: Setter<string>,
     setInvoice: Setter<string>,
     setOnchainAddress: Setter<string>,
-) => {
+): { destinationAsset: string | null; destination: string | null } => {
     const isValidForAsset = (
         asset: typeof BTC | typeof LBTC,
         address: string,
@@ -45,19 +46,19 @@ const setDestination = (
         if (isValidForAsset(BTC, destination)) {
             setAssetReceive(BTC);
             setOnchainAddress(destination);
-            return BTC;
+            return { destinationAsset: BTC, destination };
         } else if (isValidForAsset(LBTC, destination)) {
             setAssetReceive(LBTC);
             setOnchainAddress(destination);
-            return LBTC;
+            return { destinationAsset: LBTC, destination };
         } else {
             setAssetReceive(LN);
             setInvoice(destination);
-            return LN;
+            return { destinationAsset: LN, destination };
         }
     }
 
-    return undefined;
+    return { destinationAsset: null, destination: null };
 };
 
 const isValidAsset = (asset: string) =>
@@ -84,7 +85,7 @@ const handleUrlParams = (
     setSendAmount: Setter<BigNumber>,
     setReceiveAmount: Setter<BigNumber>,
 ) => {
-    const destinationAsset = setDestination(
+    const { destinationAsset, destination } = setDestination(
         setAssetReceive,
         setInvoice,
         setOnchainAddress,
@@ -103,8 +104,8 @@ const handleUrlParams = (
         }
     }
 
-    // Lightning invoice amounts take precedence
-    if (destinationAsset !== LN) {
+    // Lightning invoice amounts take precedence unless this is a LN addr
+    if (destinationAsset !== LN || isLnurl(destination)) {
         const sendAmount = parseAmount(getUrlParam("sendAmount"));
         if (sendAmount) {
             setAmountChanged(Side.Send);
--- a/src/components/CreateButton.tsx
+++ b/src/components/CreateButton.tsx
@@ -341,7 +341,8 @@ const CreateButton = () => {
             swapType() === SwapType.Submarine &&
             (lnurl() !== "" || bolt12Offer() !== undefined) &&
             amountValid() &&
-            sendAmount().isGreaterThan(0)
+            sendAmount().isGreaterThan(0) &&
+            assetReceive() !== assetSend()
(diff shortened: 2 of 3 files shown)


djkazic:
Pushed some CI fixes.

kilrau (changes requested):
Can't get this PR to set a `sendAmount` or `receiveAmount` in conjunction with a LNURL/Lightningaddress.

E.g. https://localhost:5173/?destination=kilrau@getalby.com&sendAmount=2000&sendAsset=L-BTC

djkazic:
I'll take a look at this again today.

src/context/Create.tsx
```diff
@@ -83,6 +85,7 @@ const handleUrlParams = (
     setAmountChanged: Setter<Side>,
     setSendAmount: Setter<BigNumber>,
     setReceiveAmount: Setter<BigNumber>,
+    invoice: Accessor<string>,
 ) => {
     const destinationAsset = setDestination(
```
kiocosta, line 90: What if we change the return of `setDestination` to this:
```suggestion
    const { asset: destinationAsset, destination } = setDestination(
```

```
const setDestination = {
    setAssetReceive: Setter<string>,
    setInvoice: Setter<string>,
    setOnchainAddress: Setter<string>,
): { asset: string | null; destination: string | null } => {
    ...
    ...
    if (urlParamIsSet(destination)) {
        if (isValidForAsset(BTC, destination)) {
            setAssetReceive(BTC);
            setOnchainAddress(destination);
            return { asset: BTC, destination };
        } else if (isValidForAsset(LBTC, destination)) {
            setAssetReceive(LBTC);
            setOnchainAddress(destination);
            return { asset: LBTC, destination };
        } else {
            setAssetReceive(LN);
            setInvoice(destination);
            return { asset: LN, destination };
        }
    }

    return { asset: null, destination: null };
};
```
You'd then be able to use it here:
```
if (destinationAsset !== LN || isLnurl(destination)) {
```
This would eliminate the need of passing `invoice` to `handleUrlParams` and also the need for `untrack`. What do you think?

kiocosta (commented):
Tested ACK. Also tested the below scenarios and everything worked:
```
?sendAmount=100000&destination=test@lnurl.com&sendAsset=BTC
?sendAmount=100000&destination=bcrt1qgzzhsnvstqjd5nyr0u6fz706up6ap0jy4ajs3x&sendAsset=LN
?sendAmount=100000&destination=el1qqd60krw3zqwdg97gs4skuqjf05lgf6jht3w8fsg0hyhtyv203yqnjuhnd96t0an6nz8e35n0ndqh0rvmaq5g7dj999pl26nwr&sendAsset=BTC

File: src/context/Create.tsx (hunk 1 above)
Comment (djkazic) [reply src/context/Create.tsx:90]: Yeah I like this approach. I'm not really a big TypeScript user so my initial approach was a bit hamfisted.

Comment (djkazic) [reply src/context/Create.tsx:90]: Refactored and force pushed.

**Comment by djkazic** [comment]
> Tested the below scenarios and everything worked:
> 
> ```
> ?destination=kilrau@getalby.com&sendAmount=2000&sendAsset=L-BTC
> ?sendAmount=100000&destination=test@lnurl.com&sendAsset=BTC
> ?sendAmount=100000&destination=bcrt1qgzzhsnvstqjd5nyr0u6fz706up6ap0jy4ajs3x&sendAsset=LN
> ?sendAmount=100000&destination=el1qqd60krw3zqwdg97gs4skuqjf05lgf6jht3w8fsg0hyhtyv203yqnjuhnd96t0an6nz8e35n0ndqh0rvmaq5g7dj999pl26nwr&sendAsset=BTC
> ```
> 
> One minor thing I noticed:
> 
> When I set both **assetSend** and **assetReceive** to BTC, it correctly shows the `Create` button disabled: image `?sendAmount=100000&destination=bcrt1qgzzhsnvstqjd5nyr0u6fz706up6ap0jy4ajs3x&sendAsset=BTC`
> 
> However, the button keeps enabled when I do the same with LN: image `?sendAmount=100000&destination=test@lnurl.com&sendAsset=LN`
> 
> Clicking the button results in: image

Hmm, I'm not sure why this is the case. I can take a look at it tomorrow.

**Comment by kiocosta** [comment]
> Hmm, I'm not sure why this is the case. I can take a look at it tomorrow.

The problem seems to be in `validWayToFetchInvoice` from the `CreateButton`:
```
     <button
            id="create-swap-button"
            data-testid="create-swap-button"
            class={buttonClass()}
            disabled={
                !online() ||
                !(valid() || validWayToFetchInvoice()) ||
                buttonDisable()
            }
            onClick={buttonClick}>
            {getButtonLabel(buttonLabel())}
        </button>
```
I think you also need to add `&& assetReceive() !== assetSend()` to this condition, but please confirm:
```
    const validWayToFetchInvoice = (): boolean => {
        return (
            swapType() === SwapType.Submarine &&
            (lnurl() !== "" || bolt12Offer() !== undefined) &&
            amountValid() &&
            sendAmount().isGreaterThan(0)
        );
    };
```

kilrau:
> > Hmm, I'm not sure why this is the case. I can take a look at it tomorrow.
> 
> The problem seems to be in `validWayToFetchInvoice` from the `CreateButton`:
> 
> ```
>      <button
>             id="create-swap-button"
>             data-testid="create-swap-button"
>             class={buttonClass()}
>             disabled={
>                 !online() ||
>                 !(valid() || validWayToFetchInvoice()) ||
>                 buttonDisable()
>             }
>             onClick={buttonClick}>
>             {getButtonLabel(buttonLabel())}
>         </button>
> ```
> 
> I think you also need to add `&& assetReceive() !== assetSend()` to this condition, but please confirm:
> 
> ```
>     const validWayToFetchInvoice = (): boolean => {
>         return (
>             swapType() === SwapType.Submarine &&
>             (lnurl() !== "" || bolt12Offer() !== undefined) &&
>             amountValid() &&
>             sendAmount().isGreaterThan(0)
>         );
>     };
> ```

separate issue

kiocosta (approved):
tACK

michael1011 (approved):
Thank you!

Issue #856, opened 2025-03-17
Add Refund Flow tests


kilrau:
>The refund flow could still use some additional automated tests. Haven't added them for now because this PR is already pretty big.

_Originally posted by @kiocosta in https://github.com/BoltzExchange/boltz-web-app/issues/828#issuecomment-2727687681_

michael1011:
This needs more context. We have tests for the refund flow. What's missing is the coded added in #828

kiocosta:
@kilrau assign it to me, please.

PR #864, opened 2025-03-25
feat: Refund all UTXOs in one transaction

### LINKED ISSUE #855: Refund all UTXOs in one transaction
To be done after #828

https://github.com/BoltzExchange/boltz-web-app/blob/2060d80ef1f8e012affa70132fdc5f5db5a77f1a/src/components/RefundButton.tsx#L170
Iterating over them and refunding individually cannot be shiped

kiocosta:
Closes #855 

Depends on #828

### DIFF
--- a/src/utils/refund.ts
+++ b/src/utils/refund.ts
@@ -38,12 +38,12 @@ import { createMusig, hashForWitnessV1, tweakMusig } from "./taproot/musig";
 
 const refundTaproot = async <T extends TransactionInterface>(
     swap: SubmarineSwap | ChainSwap,
-    lockupTx: TransactionInterface,
+    lockupTxs: TransactionInterface[],
     privateKey: ECPairInterface,
     decodedAddress: DecodedAddress,
     feePerVbyte: number,
-    timeoutBlockHeight: number,
     cooperative: boolean = true,
+    timeoutBlockHeight?: number,
     // Keep the error of the cooperative refund to show a nice error reason in case the
     // uncooperative transaction is not ready to be broadcast yet
     cooperativeError?: string,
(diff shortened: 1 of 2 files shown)


kilrau (changes requested):
Imported a mainnet backup with a history of old swaps and sent multiple UTXOs to an old swap's lockup address.

Refunding fails with
Fails with `logs.ts:67 Cannot read properties of undefined (reading 'timeoutBlockHeight')`.
image

just out of interest I ran the same backup with PR 282 and here refund fails with `invalid mnemonic type: undefined`
image

kiocosta:
> Refunding fails with Fails with `logs.ts:67 Cannot read properties of undefined (reading 'timeoutBlockHeight')`. ![image]

I see the problem, pushed the fix. Will make sure to include a similar test scenario on #856

> just out of interest I ran the same backup with PR 282 and here refund fails with `invalid mnemonic type: undefined`

You mean #828, right?
I have the impression it's unrelated to #828, but will try to reproduce nonetheless. 
If you use the same swap on this PR I'm guessing you will have the same issue you just had on #828.

kilrau:
Yes, 828 - mistyped. Same swap, ignore it.

src/utils/refund.ts
```diff
@@ -105,28 +104,32 @@ const refundTaproot = async <T extends TransactionInterface>(
...
-            refundTx,
-            0,
-        );
-        musig.aggregateNonces([[boltzPublicKey, boltzSig.pubNonce]]);
-        musig.initializeSession(
-            hashForWitnessV1(
-                swap.assetSend,
-                getNetwork(swap.assetSend),
-                details,
+        for (const [index, input] of refundTx.ins.entries()) {
```
kilrau, line 107: So you are looping over the refundable inputs and create refund txes individually

kilrau (approved):
tACK

~Note: refunding two UTXO in lockup in two separate transactions which is perfectly fine: ~
image


~Please rename PR and commit "in one transaction" is misleading.~

src/utils/refund.ts
kiocosta: @kilrau no, this is just to sign the multiple inputs of a single transaction. I only return one tx after this:
```
        return {
            transaction: refundTx as T,
        };
```

michael1011 (approved):

PR #868, opened 2025-03-27
chore: bump lnurl & bolt12 resolve timeout to 25 seconds

kilrau:
First resolve reliably takes up to 18 seconds for me. Thereafter it's much faster...

### DIFF
--- a/src/components/CreateButton.tsx
+++ b/src/components/CreateButton.tsx
@@ -35,6 +35,9 @@ import {
 } from "../utils/swapCreator";
 import { validateResponse } from "../utils/validation";
 
+const LNURL_FETCH_TIMEOUT = 25_000;
+const BIP353_FETCH_TIMEOUT = 25_000;
+
 export const getClaimAddress = async (
     assetReceive: Accessor<string>,
     signer: Accessor<Signer>,
@@ -354,7 +357,7 @@ const CreateButton = () => {
                     return new Promise<string>(async (resolve, reject) => {
                         const timeout = setTimeout(
                             () => reject(new Error(t("timeout"))),
-                            5_000,
+                            LNURL_FETCH_TIMEOUT,
                         );
 
                         try {
@@ -375,7 +378,7 @@ const CreateButton = () => {
                     return new Promise<string>(async (resolve, reject) => {
                         const timeout = setTimeout(
                             () => reject(new Error(t("timeout"))),
-                            15_000,
+                            BIP353_FETCH_TIMEOUT,
                         );
 
                         try {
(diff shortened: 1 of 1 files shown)


src/components/CreateButton.tsx
```diff
@@ -375,7 +375,7 @@ const CreateButton = () => {
                     return new Promise<string>(async (resolve, reject) => {
                         const timeout = setTimeout(
                             () => reject(new Error(t("timeout"))),
-                            15_000,
+                            20_000,
```
michael1011, line 378: If it takes up to 18 seconds for you, let's add some more buffer and set it to 25 seconds
Also, let's put it in a constant

michael1011 (changes requested):

src/components/CreateButton.tsx
michael1011: Maybe we should bump LNURL fetch timeouts the same way. If a request resolves successfuly, that result will be used immediately. No harm in using the same high timeout for both

src/components/CreateButton.tsx
```diff
@@ -35,6 +35,9 @@ import {
 } from "../utils/swapCreator";
 import { validateResponse } from "../utils/validation";
 
+const LNURL_FETCH_TIMEOUT = 25_000;
+const BIP353_FETCH_TIMEOUT = 25_000;
```
michael1011, line 39:
```suggestion
// In seconds
const invoiceFetchTimeout = 25_000;
```

src/components/CreateButton.tsx
```diff
@@ -354,7 +357,7 @@ const CreateButton = () => {
                     return new Promise<string>(async (resolve, reject) => {
                         const timeout = setTimeout(
                             () => reject(new Error(t("timeout"))),
-                            5_000,
+                            LNURL_FETCH_TIMEOUT,
```
michael1011, line 360:
```suggestion
                            invoiceFetchTimeout,
```

src/components/CreateButton.tsx
```diff
@@ -375,7 +378,7 @@ const CreateButton = () => {
                     return new Promise<string>(async (resolve, reject) => {
                         const timeout = setTimeout(
                             () => reject(new Error(t("timeout"))),
-                            15_000,
+                            BIP353_FETCH_TIMEOUT,
```
michael1011, line 381:
```suggestion
                            invoiceFetchTimeout,
```

michael1011 (changes requested):

michael1011 (approved):

PR #871, opened 2025-03-29
feat: add loading signal to `ButtonCreate`

### LINKED ISSUE #201: Loading Animation
First load on slow connections like with TOR can take a while, let's have a very simple (needs to be super light to load) loading animation with our logo slowly beating, compare https://opendex.app/

image

### LINKED ISSUE #817: Loading animation for resolving invoice and creating swap
API calls via Tor can take several seconds to resolve e.g. a Lightning address and then call Boltz API. 15-20s is not a stranger. We need a loading animation here, ideally in the button. I think we have our 3 dot loading animation in a button somewhere already, but can't recall where.
Image

kiocosta:
Closes #817

image

### DIFF
--- a/src/components/CreateButton.tsx
+++ b/src/components/CreateButton.tsx
@@ -34,6 +34,7 @@ import {
     createSubmarine,
 } from "../utils/swapCreator";
 import { validateResponse } from "../utils/validation";
+import LoadingSpinner from "./LoadingSpinner";
 
 // In milliseconds
 const invoiceFetchTimeout = 25_000;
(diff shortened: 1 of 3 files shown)


src/style/index.scss
```diff
@@ -743,3 +743,12 @@ header {
...
 }
+
+.inner-loading {
+    margin: 0;
+
+    & div {
+        width: 14px;
+        height: 14px;
+    }
+}
```
michael1011, line 754: Imho should go into `loadingSpinner.scss`

michael1011 (changes requested):
Can be merged after

src/style/index.scss
kiocosta: @michael1011 indeed, done!

michael1011 (approved):

kilrau:
Feels much better with the animation, noice!

PR #873, opened 2025-03-30
refactor: race invoice timeout

michael1011:
To avoid having to wait for both, LNURL and BIP-353, promises to resolve


kiocosta (approved):
tACK.

Tested with `ln@bol.tz` (lnurl) and `matt@mattcorallo.com` (BIP 353).

Also simulated scenarios where one of the requests took longer than expected by adding some delay to it, and everything worked as expected.

PR #880, opened 2025-04-01
Get python3 location from env in shebang

DanGould:
Rather than require python3 be installed exactly at /bin/python3


michael1011 (approved):
Thank you!

PR #883, opened 2025-04-04
feat: Use compiled network configuration file

### LINKED ISSUE #878: Move from config.json to have config values compile in web app
@kiocosta

kiocosta:
Closes #878
Migrating from dynamically loaded network configuration files to compiled files as a way of preventing cache-related issues.

- I decreased the scope of the test `isInvoice` inside `invoice.spec.ts` because network configuration is now static. AFAIK, it's not possible to execute a script to replace the file during the execution of this test.


.github/workflows/CI.yml
```diff
@@ -16,6 +16,8 @@ jobs:
 
       - run: npm ci
 
+      - run: npm run regtest
+
```
michael1011, line 20: Let's run this run this in `postinstall`:

```
"postinstall": "npm run regtest"
```

src/configs/beta.ts
```diff
@@ -0,0 +1,73 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+/* @ts-ignore */
+import { baseConfig, chooseUrl } from "./configs/base";
```
michael1011, line 3: We can avoid this by using absolute imports

```suggestion
import { baseConfig, chooseUrl } from "src/configs/base";
```

src/configs/beta.ts
```diff
@@ -0,0 +1,73 @@
...
+            rifRelay: "https://boltz.mainnet.relay.rifcomputing.net",
+            contracts: {
+                deployHeight: 6747215,
+                smartWalletFactory:
+                    "0x44944a80861120B58cc48B066d57cDAf5eC213dd",
+                deployVerifier: "0xc0F5bEF6b20Be41174F826684c663a8635c6A081",
+            },
+        },
+    },
+};
```
michael1011, line 71: To make sure that the configs adhere to the type

```suggestion
} as Config;
```

src/configs/beta.ts
```diff
@@ -0,0 +1,73 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+/* @ts-ignore */
+import { baseConfig, chooseUrl } from "./configs/base";
+
+const config = {
+    ...baseConfig,
+    isBeta: true,
```
michael1011, line 7: The beta config is the mainnet config with `isBeta: true`, so we should import the mainnet config and inherit the values just like we do with the `baseConfig`

src/configs/pro.ts
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+/* @ts-ignore */
+import { baseConfig, chooseUrl } from "./configs/base";
+
+const config = {
+    ...baseConfig,
+    isPro: true,
```
michael1011, line 7: Same thing as with beta, just with `isPro: true`

michael1011 (changes requested):

src/configs/beta.ts
kiocosta: @michael1011 added it, also had to define this path on `vite.config.mjs`:
```
    resolve: {
        alias: {
            src: path.resolve(__dirname, "src"),
        },
    },
```
Opted to define just this one as we usually don't use aliases.

kiocosta:
@michael1011 pushed new changes addressing the requested changes.
I decided to also add an error message in case a dev tries to start the server without a config file:
```
$ npm run dev

> boltz-frontend@1.7.2 dev
> vite

❌ Missing configuration file: src/config.ts

Please run one of the following commands to generate a config file:
    - npm run mainnet
    - npm run regtest
    - npm run testnet
    - npm run beta
    - npm run pro

Then start the dev server again.
```

michael1011 (approved):

PR #885, opened 2025-04-05
fix: urlParam `receiveAsset` not working

### LINKED ISSUE #884: URL Params: setting `receiveAsset` doesn't work anymore
`receiveAsset` is ignored, test e.g. with https://boltz.exchange/?sendAsset=BTC&receiveAsset=RBTC . Likely that https://github.com/BoltzExchange/boltz-web-app/blob/main/src%2Fcontext%2FCreate.tsx#L100-L101 bricked it. Let's fix it and add tests for all URL params.

kiocosta:
Closes #884

The problem occurred when`setDestination` returned `destinationAsset` as `null`, but the `if` checked if `destinationAsset` was `undefined`.
https://github.com/BoltzExchange/boltz-web-app/blob/cb62f517c97cf35e798922b79feaa8900cd88f16/src/context/Create.tsx#L88-L100

Fixed by returning `destinationAsset` as `undefined` (instead of redefining it as `null`) and added unit tests covering this scenario.

### DIFF
--- a/src/context/Create.tsx
+++ b/src/context/Create.tsx
@@ -23,7 +23,7 @@ const setDestination = (
     setAssetReceive: Setter<string>,
     setInvoice: Setter<string>,
     setOnchainAddress: Setter<string>,
-): { destinationAsset: string | null; destination: string | null } => {
+): { destinationAsset?: string; destination?: string } => {
     const isValidForAsset = (
         asset: typeof BTC | typeof LBTC,
         address: string,
@@ -58,7 +58,7 @@ const setDestination = (
         }
     }
 
-    return { destinationAsset: null, destination: null };
+    return { destinationAsset: undefined, destination: undefined };
 };
 
 const isValidAsset = (asset: string) =>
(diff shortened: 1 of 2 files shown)


kilrau (approved):
tACK

src/context/Create.tsx
```diff
@@ -23,7 +23,7 @@ const setDestination = (
     setAssetReceive: Setter<string>,
     setInvoice: Setter<string>,
     setOnchainAddress: Setter<string>,
-): { destinationAsset: string | null; destination: string | null } => {
+): { destinationAsset: string; destination: string } => {
```
michael1011, line 26:
```suggestion
): { destinationAsset?: string; destination?: string } => {
```

michael1011 (changes requested):
Works but needs the correct return type

src/context/Create.tsx
kiocosta: @michael1011 fixed :)

michael1011 (approved):

PR #886, opened 2025-04-06
feat: improve UX for copying address/invoice on click

### LINKED ISSUE #875: Copy Invoice/Address on Click
Had two users recently that didn't notice or understand our "Copy Invoice" button on Desktop.
Image

Let's add a "[Click to copy](https://masteringjs.io/tutorials/fundamentals/copy-to-clipboard)" function to the invoice field. Right now one can only mark and copy which copies the invoice with `...` which is the last thing we want.

Image

Let's do the same for addresses in submarine and chain swaps:
Image

kiocosta:
Closes #875

The issue was about adding a "copy on click" functionality to the address/invoice "input", however the functionality was already there, it just wasn't clear due to lack of clues about it.

This PR adds to the "input" the same icon aspect & behavior we already had in other buttons:
image
image

### DIFF
--- a/src/components/CopyBox.tsx
+++ b/src/components/CopyBox.tsx
@@ -1,0 +1,33 @@
+import { BiRegularCopy } from "solid-icons/bi";
+import { IoCheckmark } from "solid-icons/io";
+import { Show, createSignal } from "solid-js";
+
+import { copyIconTimeout } from "../consts/CopyContent";
+import { clipboard, cropString, isMobile } from "../utils/helper";
+
+const CopyBox = (props: { value: string }) => {
+    const [copyBoxActive, setCopyBoxActive] = createSignal(false);
+    const baseStrLength = isMobile() ? 8 : 13;
+    const maxStrLength = baseStrLength + 3; // 3 is the length of the ellipsis
+
+    const copyBoxText = () => {
+        clipboard(props.value);
+        setCopyBoxActive(true);
+        setTimeout(() => {
+            setCopyBoxActive(false);
+        }, copyIconTimeout);
+    };
+
+    return (
+        <p onClick={copyBoxText} class="copy-box break-word">
+            <Show
+                when={copyBoxActive()}
+                fallback={<BiRegularCopy size={23} data-testid="copy-icon" />}>
+                <IoCheckmark size={23} data-testid="checkmark-icon" />
+            </Show>
+            {cropString(props.value, baseStrLength, maxStrLength)}
+        </p>
+    );
+};
+
+export default CopyBox;
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -36,10 +36,7 @@ export default [
     {
         rules: {
             "no-async-promise-executor": "off",
-        },
-    },
-    {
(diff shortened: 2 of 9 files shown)


kilrau (changes requested):
Prevent selection of invoice text works, the icon and tick animation is nice and subtle 👍

Just that the icon currently breaks the UI depending on invoice char length (try with liquid swap, looks like here invoices are at least 1 char longer):
image

And screen resolution (lower bar is iPhone SE), either dynamically shorten more on smaller resolutions to always make it fit in one line or center the icon vertically:
image

kiocosta:
>Just that the icon currently breaks the UI depending on invoice char length (try with liquid swap, looks like here invoices are at least 1 char longer):

Should be good now! Watch-out that I am using "isMobile()" to adjust the length of the string; you need to press F5 when you switch between mobile and desktop version so this function runs again and renders accordingly

kilrau (approved):
LGTM

src/components/CopyBox.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const CopyBox = (props: { value: string }) => {
+    const [copyBoxActive, setCopyBoxActive] = createSignal(false);
+
+    const copyBoxText = () => {
+        clipboard(props.value);
+        setCopyBoxActive(true);
+        setTimeout(() => {
+            setCopyBoxActive(false);
+        }, 600);
```
michael1011, line 15: Let's use a constant for `600` and use it in `CopyButton.tsx` too to ensure consistency

michael1011 (changes requested):

src/components/CopyBox.tsx
kiocosta: @michael1011 created a `CopyContent.ts` under `/consts` and exported it from there

src/components/CopyBox.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const [copyBoxActive, setCopyBoxActive] = createSignal(false);
+
+    const copyBoxText = () => {
+        clipboard(props.value);
+        setCopyBoxActive(true);
+        setTimeout(() => {
+            setCopyBoxActive(false);
+        }, copyIconTimeout);
+    };
+    console.log(isMobile());
```
michael1011, line 18:
```suggestion
```
Sorry, missed that one the first time

michael1011 (changes requested):

src/components/CopyBox.tsx
kiocosta: @michael1011 actually my bad for not seeing it either, removed

e2e/utils.ts
```diff
@@ -24,7 +25,7 @@ const execCommand = async (command: string): Promise<string> => {
 
         return stdout.trim();
     } catch (error) {
-        console.error(`Failed to execute command: ${command}`, error);
+        log.error(`Failed to execute command: ${command}`, error);
```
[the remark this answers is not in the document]
kiocosta: @michael1011  disabled the rule for e2e/ and tests/

PR #895, opened 2025-04-09
feat: prompt on tab close with pending swaps

### LINKED ISSUE #815: Browser prompt on tab close with pending reverse or chain swaps
Had a flood of support tickets these days where folks created a swap, sent funds to the lockup address or lightning invoice and closed the browser. Because ya know they felt like they are done. Happens a lot with privacy browsers like Tor, which is especially crucial as the only way to rescue funds here is a refund file refund. Users always only realized *after* my explanation why they'd need to keep the browser open. So, I want a prompt on tab close IF (and only if) there are 1. Reverse Swaps or 2. Chain Swaps ongoing, meaning in any state more advanced than `swap.created`.

Text of the closing prompt: `You have pending Swaps!` `Your swaps are likely to fail if you close this page now. Are you sure you want to close this page?`

https://www.geeksforgeeks.org/how-to-detect-browser-or-tab-closing-in-javascript/

kiocosta:
Closes #815

Draft: still need to test it more.

### DIFF
--- a/src/components/SwapChecker.tsx
+++ b/src/components/SwapChecker.tsx
@@ -1,6 +1,7 @@
 import { OutputType } from "boltz-core";
 import log from "loglevel";
 import { createEffect, onCleanup, onMount } from "solid-js";
+import { createStore } from "solid-js/store";
 
 import { BTC, LBTC, RBTC } from "../consts/Assets";
 import { SwapType } from "../consts/Enums";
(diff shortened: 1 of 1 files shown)


src/context/Global.tsx
```diff
@@ -156,6 +160,8 @@ const GlobalProvider = (props: { children: JSX.Element }) => {
 
     const [hideHero, setHideHero] = createSignal<boolean>(false);
 
+    const [pendingSwaps, setPendingSwaps] = createStore<string[]>([]);
```
michael1011, line 163: This is only used in `SwapChecker` and should be defined there

src/components/SwapChecker.tsx
```diff
@@ -268,6 +277,9 @@ export const SwapChecker = () => {
                 const claimedSwap = await getSwap(res.id);
                 claimedSwap.claimTx = res.claimTx;
                 await setSwapStorage(claimedSwap);
+                setPendingSwaps((pendingSwaps) =>
+                    pendingSwaps.filter((id: string) => id !== res.id),
+                );
```
michael1011, line 282: We have to remove from the array when we refund too. How about instead of doing this, we check if we should remove the swap from the pending list whenever we update the status of a swap?

src/context/Global.tsx
kiocosta: @michael1011 true, fixed

src/components/SwapChecker.tsx
kiocosta: @michael1011 makes sense! I've updated it to add/remove from the pending list when the swap status updates.

kiocosta:
Closed this to deploy the branch to Vercel on #897

PR #896, opened 2025-04-09
chore: reorder header menu

kilrau:
moved "help" one position to the left to have it even more visible

image

### DIFF
--- a/src/components/Nav.tsx
+++ b/src/components/Nav.tsx
@@ -72,7 +72,7 @@ const Nav = (props: { network: string; isPro?: boolean }) => {
                         {t("history")}
                     </A>
 
-                    <Show when={config.discordUrl}>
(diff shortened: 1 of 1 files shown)


kiocosta (approved):
ACK

src/components/Nav.tsx
```diff
@@ -72,21 +72,21 @@ const Nav = (props: { network: string; isPro?: boolean }) => {
                         {t("history")}
                     </A>
 
-                    <Show when={config.docsUrl}>
+                    <Show when={config.discordUrl}>
```
michael1011, line 75:
```suggestion
                    <Show when={config.supportUrl}>
```
Just noticed that this is saying `discordUrl`

michael1011 (changes requested):

PR #897, opened 2025-04-09
feat: prompt on tab close with pending swaps

### LINKED ISSUE #815: Browser prompt on tab close with pending reverse or chain swaps

kiocosta:
Closes #815 

Displays generic prompt when user has Reverse or Chain swap pending and is on desktop.
Tested on Tor browser:
image

Also displays this warning (mobile AND BTC -> x only):
image

### DIFF
--- a/src/components/SwapChecker.tsx
+++ b/src/components/SwapChecker.tsx
@@ -1,6 +1,7 @@
 import { OutputType } from "boltz-core";
 import log from "loglevel";
 import { createEffect, onCleanup, onMount } from "solid-js";
+import { createStore } from "solid-js/store";
 
 import { BTC, LBTC, RBTC } from "../consts/Assets";
 import { SwapType } from "../consts/Enums";
@@ -27,6 +28,7 @@ import { getApiUrl } from "../utils/helper";
 import type {
     ChainSwap,
     ReverseSwap,
+    SomeSwap,
     SubmarineSwap,
 } from "../utils/swapCreator";
 import { getRelevantAssetForSwap } from "../utils/swapCreator";
@@ -178,6 +180,28 @@ export const SwapChecker = () => {
 
     let ws: BoltzWebSocket | undefined = undefined;
 
+    const [pendingSwaps, setPendingSwaps] = createStore<string[]>([]);
+
+    const updatePendingSwaps = (swap: SomeSwap, data: SwapStatus) => {
+        if (![SwapType.Chain, SwapType.Reverse].includes(swap.type)) {
+            return;
+        }
+
+        if (
+            Object.values(swapStatusPending).includes(data.status) &&
+            data.status !== swapStatusPending.SwapCreated &&
+            !pendingSwaps.includes(swap.id)
+        ) {
+            setPendingSwaps((pendingSwaps) => [...pendingSwaps, swap.id]);
+        }
+
+        if (Object.values(swapStatusFinal).includes(data.status)) {
+            setPendingSwaps((pendingSwaps) =>
+                pendingSwaps.filter((id: string) => id !== swap.id),
+            );
+        }
+    };
+
     const prepareSwap = async (swapId: string, data: SwapStatus) => {
         const currentSwap = await getSwap(swapId);
         if (currentSwap === null) {
@@ -194,6 +218,7 @@ export const SwapChecker = () => {
             }
         }
         if (data.status) {
+            updatePendingSwaps(currentSwap, data);
             await updateSwapStatus(currentSwap.id, data.status);
         }
     };
@@ -373,6 +398,14 @@ export const SwapChecker = () => {
             ws.subscribeUpdates([activeSwap.id]);
         }
     });
+
+    window.onbeforeunload = (event: BeforeUnloadEvent) => {
+        if (pendingSwaps?.length > 0) {
+            event.preventDefault();
+            return "";
+        }
+        return undefined;
+    };
 
     return "";
 };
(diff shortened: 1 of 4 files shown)


src/status/TransactionMempool.tsx
```diff
@@ -1,12 +1,21 @@
...
 
-const TransactionMempool = () => {
+const TransactionMempool = (props: { swap: Accessor<SomeSwap> }) => {
     const { t } = useGlobalContext();
+
     return (
         <div>
             <h2>{t("tx_in_mempool")}</h2>
             <p>{t("tx_in_mempool_subline")}</p>
+            {isMobile() && props.swap().assetSend === BTC && (
```
michael1011, line 16: Same thing for L-BTC and RSK
Also, only for reverse chain chain swaps

src/components/SwapChecker.tsx
```diff
@@ -374,5 +395,13 @@ export const SwapChecker = () => {
         }
     });
 
+    // @ts-expect-error: onbeforeunload doesn't need to return a value
+    window.onbeforeunload = (event: BeforeUnloadEvent) => {
```
michael1011, line 399: Why not return something anyways? https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

src/components/SwapChecker.tsx
```diff
@@ -178,6 +179,23 @@ export const SwapChecker = () => {
...
 
+    const [pendingSwaps, setPendingSwaps] = createStore<string[]>([]);
+
+    const handlePendingSwap = (swapId: string, data: SwapStatus) => {
+        if (
+            Object.values(swapStatusPending).includes(data.status) &&
+            data.status !== swapStatusPending.SwapCreated &&
+            !pendingSwaps.includes(swapId)
+        ) {
+            setPendingSwaps((pendingSwaps) => [...pendingSwaps, swapId]);
```
michael1011, line 190: We should only add reverse and chain swaps here because submarines will succeed even when the tab is closed

src/status/TransactionMempool.tsx
kiocosta: @michael1011 updated it to this:
```
isMobile() &&
                [SwapType.Chain, SwapType.Reverse].includes(
                    props.swap().type,
                )
```

src/components/SwapChecker.tsx
kiocosta: @michael1011 moved the verification of the swap status to inside this function:
```
    const updatePendingSwaps = (swap: SomeSwap, data: SwapStatus) => {
        if (![SwapType.Chain, SwapType.Reverse].includes(swap.type)) {
            return;
        }
```

kiocosta:
@kilrau when you can, please add the i18n translations for this one :)

kilrau (changes requested):
`BeforeUnloadEvent` triggers even though I didn't even start a swap yet.
image
image

Also can you exclude this from reload? Not useful.
image

Displaying the warning on mobile and BTC -> X only works as expected.

kiocosta:
@kilrau pushed new changes. Fixed the bug of dialog showing up even with no pending swaps. Re-validated that it only appears when there are pending swaps.

kilrau:
Now it doesn't trigger at all for me on desktop for reverse swaps.

kilrau (approved):
Now works as expected, only gets triggered when there is a reverse (very hard to test as web app claims within a secons) and chain swap is funded and processing.

Screenshot from 2025-04-17 15-16-51

src/components/SwapChecker.tsx
```diff
@@ -27,6 +28,7 @@ import { getApiUrl } from "../utils/helper";
 import type {
     ChainSwap,
     ReverseSwap,
+    SomeSwap,
```
kilrau, line 31: That really the best name?

kilrau:
Still needs @michael1011 's review

src/components/SwapChecker.tsx
michael1011: It's either submarine, reverse or chain swap. If you come up with a better name, PRs welcome

src/status/TransactionMempool.tsx
```diff
@@ -1,12 +1,22 @@
...
     const { t } = useGlobalContext();
+
     return (
         <div>
             <h2>{t("tx_in_mempool")}</h2>
             <p>{t("tx_in_mempool_subline")}</p>
+            {isMobile() &&
+                [SwapType.Chain, SwapType.Reverse].includes(
+                    props.swap().type,
+                ) && <h3>{t("tx_in_mempool_warning")}</h3>}
```
michael1011, line 19: Let's use a `<Show>` for this

src/i18n/i18n.ts
```diff
@@ -378,6 +380,8 @@ const dict = {
         tx_in_mempool: "Transaktion befindet sich im Mempool.",
         tx_in_mempool_subline:
             "Warte auf Bestätigung, um den Swap abzuschließen.",
+        tx_in_mempool_warning:
+            "Tab nicht schließen, bis Swap abgeschlossen ist!",
```
michael1011, line 384:
```suggestion
            "Tab nicht schließen, bis der Swap abgeschlossen ist!",
```

michael1011 (changes requested):

michael1011 (approved):

PR #898, opened 2025-04-10
fix: address derivation for RSK from Trezor

michael1011:
Fixes `Deriving addresses failed: bad address checksum` for our native Trezor integration


kilrau (approved):
works

PR #900, opened 2025-04-11
fix: Rootstock derivation path

kilrau:


src/utils/hardware/HadwareSigner.ts
```diff
@@ -5,11 +5,11 @@ export const derivationPaths = {
 };
 
 export const derivationPathsMainnet = {
-    RSK: "44'/137'/0'",
+    Rootstock: "44'/137'/0'/0",
 };
 
 export const derivationPathsTestnet = {
-    ["RSK Testnet"]: "44'/37310'/0'",
+    ["Rootstock Testnet"]: "44'/37310'/0'/0",
```
michael1011, line 12: image

Causes a line break

michael1011 (changes requested):

kilrau:
Alright ser fixing that line break later when I have screen time, just wanted to leave a comment that the user in question tested with this and it works. Derives the correct Rootstock address now.

src/utils/hardware/HadwareSigner.ts
kilrau: image

michael1011 (approved):

PR #907, opened 2025-04-18
fix: missing colon for max amount

michael1011:


kilrau (approved):
My kinda PRs, good catch!

PR #908, opened 2025-04-23
fix: ensure text does not wrap in settings span

### LINKED ISSUE #899: Japanese breaks settings menu on/off buttons
Looks like 2 `ja` chars are wider than 3 latin chars:
Image
Image

Either have the button widen dynamically or switch `ja` to stick with english "on"/"off" like we do for other languages (e.g. `es`).

gabriel-cmps:
Closes #899 

This PR adds the CSS property text-wrap: nowrap; to the span selector inside the .setting class in the settings.scss file.

The goal is to prevent the text inside spans from wrapping onto multiple lines.

### Before

432250915-3c7f9a9c-8f2e-4517-9315-8b6aabd4e0b8

### After

Screenshot from 2025-04-23 20-07-34


kiocosta (commented):
Thanks for the PR! :clap: 

Just a small thing, I think we should also change this to `min-width`:
https://github.com/BoltzExchange/boltz-web-app/blob/64cc380928c04ba5b08c5f46f0d1d44a65fb7203/src/style/index.scss#L369

This way, when bigger than 28px, `width` will behave like `fit-content` and prevent any overlap between the content and the border of the button.

kiocosta (approved):
tACK.

gabriel-cmps:
> Thanks for the PR! 👏
> 
> Just a small thing, I think we should also change this to `min-width`:
> 
> https://github.com/BoltzExchange/boltz-web-app/blob/64cc380928c04ba5b08c5f46f0d1d44a65fb7203/src/style/index.scss#L369
> 
> This way, when bigger than 28px, `width` will behave like `fit-content` and prevent any overlap between the content and the border of the button.

This solution using only min-width wasn't enough. I had to proportionally increase the size of the boxes from 28px to 30px.

#### When I used only min-width, this was the result:

https://github.com/user-attachments/assets/b1ef674f-e76a-4395-b6fa-d4762e78fb7f

#### Now, after increasing the box size proportionally, it looks like this:

https://github.com/user-attachments/assets/ff6bd455-da1e-496f-ba7d-cef1a56a7732

kiocosta (approved):

michael1011 (approved):
Still a tight fit but much nicer than before

Thank you for the PR

PR #914, opened 2025-05-05
fix: revalidate amount on backspace deletion

### LINKED ISSUE #718: Deleting amount with backspace doesn't trigger revalidation
If I first enter a valid address, then any amount (can be invalid)
image

And then delete the amount with backspace, web app should detect this and switch back to
image

Instead it has "Create Atomic Swap" button active and then returns this API error:
image

gabriel-cmps:
Closes #718.

### DIFF
--- a/src/pages/Create.tsx
+++ b/src/pages/Create.tsx
@@ -82,11 +82,13 @@ const Create = () => {
         }
     };
 
-    const checkEmptyAmount = (amount: string) => {
-        if (amount === "") {
-            setReceiveAmount(BigNumber(0));
-            setSendAmount(BigNumber(0));
-        }
+    const isEmptyAmount = (amount: string): boolean => {
+        return amount === "";
+    };
+
+    const resetAmounts = () => {
+        setReceiveAmount(BigNumber(0));
+        setSendAmount(BigNumber(0));
     };
 
     const changeReceiveAmount = (evt: InputEvent) => {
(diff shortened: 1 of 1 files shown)


src/pages/Create.tsx
```diff
@@ -82,11 +82,14 @@ const Create = () => {
...
-            setSendAmount(BigNumber(0));
-        }
+    const shouldResetOnEmptyAmount = (amount: string): boolean => {
+        const isEmpty = amount === "";
+        if (!isEmpty) return false;
+
+        setReceiveAmount(BigNumber(0));
+        setSendAmount(BigNumber(0));
+        validateAmount();
+        return true;
```
kiocosta, line 92: IMO this needs more separation of concerns.

`setReceiveAmount`, `setSendAmount` and `validateAmount` could be separated into a single function and we could have a different function just for the boolean part.

src/pages/Create.tsx
```diff
@@ -95,7 +98,7 @@ const Create = () => {
             .trim()
             .replaceAll(" ", "")
             .replaceAll(",", ".");
-        checkEmptyAmount(amount);
+        if (shouldResetOnEmptyAmount(amount)) return;
```
kiocosta, line 101: After breaking the aforementioned function into 2, perhaps it could make sense to divide this into more lines?
This is just my opinion, but I think having the `{ }` makes life a bit easier when debugging.

kiocosta (approved):

src/pages/Create.tsx
gabriel-cmps: I agree, I'll do that.

src/pages/Create.tsx
gabriel-cmps: I hadn't thought of it that way, it actually seems better. I'll make that change.

kiocosta (approved):
Tested ACK.

michael1011 (approved):

PR #920, opened 2025-05-21
feat: add alby as integration, remove cln, move to alphabetical order

kilrau:
image


src/pages/Hero.tsx
```diff
@@ -122,59 +122,58 @@ export const Hero = () => {
                 <div id="integrations">
                     <div>
                         <a
-                            href="https://corelightning.org/"
+                            href="https://getalby.com/"
                             target="_blank"
-                            class="cln"
```
michael1011, line 127: Why remove CLN?

kilrau: Because this section should focus on high quality/prod-grade integrations. Also, technically we are not integrated with CLN directly, it's more [clboss](https://github.com/ZmnSCPxj/clboss) and some more unmaintained plugins like [nloop](https://github.com/bitbankinc/NLoop).

michael1011 (approved):

PR #930, opened 2025-06-06
Add Pro swap opportunities list

kiocos:


kilrau (changes requested):
Looks like you bricked regular Boltz with this. Updated EN strings pushed. Will add translations once the PR is merge-ready. Let's ship this!

src/components/Accordion.tsx
```diff
@@ -0,0 +1,33 @@
...
+            <button
+                class="accordion-header"
+                style={{
+                    cursor: props.onClick ? "pointer" : "default",
+                }}
+                onClick={() => props.onClick?.()}
+                aria-expanded={props.isOpen}>
+                <span class="accordion-icon">
+                    {props.isOpen ? <VsChevronDown /> : <VsChevronRight />}
+                </span>
```
michael1011, line 22: Let's hide this when there is no `onClick`

src/components/Fees.tsx
```diff
@@ -194,8 +223,14 @@ const Fees = () => {
                     />
                 </span>
                 <br />
-                {t("fee")} ({boltzFee().toString().replaceAll(".", separator())}
-                %):{" "}
+                {t("fee")} (
+                <span
+                    class={`${
+                        boltzFee() < 0 && shouldHighlightFee() && "negative-fee"
+                    } ${boltzFee() > 0 && shouldHighlightFee() && "lower-fee"}`}>
```
michael1011, line 230: Let's create a function for this to make sure we have the same logic in the comparison table

src/context/Global.tsx
```diff
@@ -294,6 +300,22 @@ const GlobalProvider = (props: { children: JSX.Element }) => {
...
     };
 
+    const fetchRegularPairs = async () => {
+        try {
+            const data = await getPairs({
+                headers: {
+                    referral: isMobile()
+                        ? "boltz_webapp_mobile"
+                        : "boltz_webapp_desktop",
+                },
```
michael1011, line 310: You can remove that

src/context/Global.tsx
```diff
@@ -294,6 +300,22 @@ const GlobalProvider = (props: { children: JSX.Element }) => {
...
+            });
+            log.debug("getalternativepairs", data);
+            setRegularPairs(data);
+        } catch (error) {
+            log.debug(error);
```
michael1011, line 315: This should throw and log an error

src/context/Global.tsx
```diff
@@ -294,6 +300,22 @@ const GlobalProvider = (props: { children: JSX.Element }) => {
...
+    const fetchRegularPairs = async () => {
+        try {
+            const data = await getPairs({
+                headers: {
+                    referral: isMobile()
+                        ? "boltz_webapp_mobile"
+                        : "boltz_webapp_desktop",
+                },
+            });
+            log.debug("getalternativepairs", data);
```
michael1011, line 312:
```suggestion
            log.debug("Regular pairs", data);
```

src/i18n/i18n.ts
```diff
@@ -574,6 +586,19 @@ const dict = {
             "Bitte gib deinen Rettungsschlüssel unten ein, um ihn zu importieren.",
         hint_paste_mnemonic:
             "Tipp: du kannst alle 12 Wörter gleichzeitig einfügen.",
+        swap_opportunities: "Verfügbare Pro-Optionen",
+        swap_opportunities_subline:
+            "Verdiene Sats oder erhalte einen Rabatt für diese Paare",
+        pro_fee: "Progebühr",
+        regular_fee: "Standardgebühr",
```
michael1011, line 593:
```suggestion
        regular_fee: "Normale Gebühr",
```
Sounds nicer imho

src/pages/Create.tsx
```diff
@@ -368,6 +388,31 @@ const Create = () => {
...
+                                if (
+                                    assetSend() !== opportunity.assetSend ||
+                                    assetReceive() !== opportunity.assetReceive
+                                ) {
+                                    setAssetSend(opportunity.assetSend);
+                                    setAssetReceive(opportunity.assetReceive);
+                                    notify(
+                                        "success",
+                                        t("swap_opportunity_selected"),
+                                    );
```
michael1011, line 409: I *really* don't think we need a notification to confirm a click

src/pages/FeeComparison.tsx
```diff
@@ -0,0 +1,45 @@
...
+    const { t, pairs, fetchPairs, regularPairs, fetchRegularPairs } =
+        useGlobalContext();
+    const navigate = useNavigate();
+
+    void fetchPairs();
+
+    onMount(() => {
+        if (config.isPro) {
+            void fetchRegularPairs();
+        }
```
michael1011, line 21: There shouldn't be scenario where we render this if it's not pro

src/style/index.scss
```diff
@@ -951,3 +951,135 @@ header {
...
+    @keyframes slideDown {
+        from {
+            opacity: 0;
+            transform: translateY(-5px);
+        }
+        to {
+            opacity: 1;
+            transform: translateY(0);
+        }
+    }
```
michael1011, line 1084: Nice!

michael1011 (changes requested):

src/components/Fees.tsx
kiocos: @michael1011 done!

src/context/Global.tsx
kiocos: @michael1011 I'm not sure if I can.
The `fetcher` function uses this:
```
    const referral = localStorage.getItem(referralIdKey) || defaultReferral();
```
If I remove these headers, it will fallback to `defaultReferral`, which in this case would be `pro`, so it would not get the regular fee.

src/pages/FeeComparison.tsx
kiocos: indeed, removed the `if`

PR #932, opened 2025-06-07
Add missing theme CSS variables

kiocos:


michael1011 (commented):
The new button hover is the same color is the disabled state. Idk about that

kiocos:
> The new button hover is the same color is the disabled state. Idk about that

Indeed, just chaged the color.
Hover:
image

Disabled:
image

michael1011 (approved):

PR #937, opened 2025-06-10
Refund improvements

### LINKED ISSUE #929: Refund address rejected with `parsing refund address failed Error: null has no matching Script`
Edge case that happens when lockup comes in after swap expired. Web app won't let user paste refund destination address in text field, logs show `"parsing refund address failed Error: null has no matching Script"`.

Definitely a web app bug, on backend side we are signing coo refunds for these states.

As @michael1011 noted probably related to the introduction of block explorer checks:
> the error is kinda misleading i think
> it's fetching from block explorer issue

kiocos:
Closes #929 

### Before
image
- Confusing disabled "Enter BTC address" input

### Now
image
- Disabled input hidden
- Added message informing possibility of a refund

Since this scenario is an edge case, I decided not to add complexity to the code for something users can handle with a simple browser refresh. We can implement a more robust solution in the future if needed, but it doesn’t seem necessary right now.

Also did a small refactor on input validation.


michael1011 (commented):
Concept ACK. Needs CI fixes and translations

michael1011 (approved):

PR #959, opened 2025-06-30
fix: Allow pasting amounts with white space on `Create`

### LINKED ISSUE #950: Allow pasting amounts with spaces in swap box
Web app should be able to handle pasted amounts that include spaces, since the amounts we display use spaces as thousand separators and one can't copy amounts from one input box to another.

https://github.com/user-attachments/assets/e3d8a4aa-a94d-49b1-a9c9-5309f3fe1052

kiocos:
Closes #950 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input validation when pasting text by removing all whitespace, ensuring only valid, space-free entries are accepted.
* **Tests**
  * Added tests to confirm correct handling of pasted amounts containing spaces and commas.

### DIFF
--- a/src/pages/Create.tsx
+++ b/src/pages/Create.tsx
@@ -176,7 +176,10 @@ const Create = () => {
 
     const validatePaste = (evt: ClipboardEvent) => {
         const clipboardData = evt.clipboardData || globalThis.clipboardData;
-        const pastedData = clipboardData.getData("Text").trim();
+        const pastedData = clipboardData
+            .getData("Text")
+            .replace(/\s+/g, "")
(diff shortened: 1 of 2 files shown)


michael1011 (changes requested):
Needs tests to avoid regressions

kiocos:
@michael1011 done.

`onPaste` wasn't being triggered by the tests we had on `Create.spec.tsx`, so I also added the paste event to one of the existing tests, because `fireEvent.input` does not trigger `onPaste`.

michael1011 (approved):

PR #972, opened 2025-07-24
chore: add "do not close this page" to mainchain submarine/chain swap…

kilrau:
Having one or two support cases about this every week. Let's see if this helps at least a lil bit. Also switched 24h-> 12h.

Before and after:
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated warning messages during swaps to clarify that the page must remain open for the swap to complete, across all supported languages.
  * Improved Portuguese translation for the transaction status message.

* **Refactor**
  * Simplified the swap warning display logic for a more consistent user experience.

### DIFF
--- a/src/status/TransactionConfirmed.tsx
+++ b/src/status/TransactionConfirmed.tsx
@@ -88,9 +88,6 @@ const TransactionConfirmed = () => {
                 <div>
                     <h2>{t("tx_confirmed")}</h2>
                     <p>{t("tx_ready_to_claim")}</p>
-                    <Show when={[SwapType.Chain].includes(swap().type)}>
-                        <h3>{t("tx_in_mempool_warning")}</h3>
-                    </Show>
                     <LoadingSpinner />
                 </div>
             }>
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -125,7 +125,7 @@ const dict = {
         tx_in_mempool: "Transaction is in mempool",
         tx_in_mempool_subline: "Waiting for confirmation to complete the swap.",
         tx_in_mempool_warning:
-            "Please keep this tab open until the swap completed!",
+            "Keep this page open, otherwise your swap can't complete!",
         expired: "Swap expired!",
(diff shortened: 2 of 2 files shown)


kiocos:
Do you know when users typically close the webapp?
I suspect they don't actually close at this stage, but only after they see the loading animation "waiting for confirmation". If this is the case, I guess adding a warning at that stage could be more effective, especially because we have less going on.

If not, maybe we can still separate this new text into two sentences to make it look cleaner. I can give it a try implementing one of these (or both) suggestions if you want.

kiocos:
@kilrau how about this?

The title "Transaction is in mempool" is redundant with the status, so I replaced it with the "Please keep this tab open" message. 

Before
image

After
image

image

I noticed we actually did something similar before, but for mobile browsers only: https://github.com/BoltzExchange/boltz-web-app/pull/897

src/status/TransactionConfirmed.tsx
```diff
@@ -88,6 +88,12 @@ const TransactionConfirmed = () => {
                 <div>
                     <h2>{t("tx_confirmed")}</h2>
                     <p>{t("tx_ready_to_claim")}</p>
+                    <Show
+                        when={[SwapType.Chain, SwapType.Reverse].includes(
+                            swap().type,
+                        )}>
+                        <h3>{t("tx_in_mempool_warning")}</h3>
+                    </Show>
```
michael1011, line 96: Why should we show this warning here too? Claiming swaps should be almost instant

kilrau: True, but also doesn't matter then doesn't it? Maybe even nicer too keep the warning to avoid a visual "jump" of text when the state switches to `Claiming...`

kilrau: Ah you mean for Reverse Swaps. Good point. Useful for https://github.com/BoltzExchange/boltz-web-app/issues/346, but not when user accepts 0-conf.

michael1011 (approved):

PR #973, opened 2025-07-24
feat: Claim pending chain/reverse swaps via rescue key rescan

kiocos:
Closes #852 
~Depends on [#1010](https://github.com/BoltzExchange/boltz-backend/pull/1010)~

WIP:
- [x] Add e2e tests
- [x] CI fix

## Summary by CodeRabbit

* **New Features**
  * Introduced a new "Rescue" section replacing "Refund" for swap recovery, including dedicated routes and navigation.
  * Added the ability to claim pending swaps via rescue key, with a new claim page and supporting UI.
  * Enhanced localization with new "rescue" and "claim" terminology and messages across all supported languages.

* **Improvements**
  * Refined swap recovery logic to distinguish between claimable, refundable, pending, and non-actionable swaps.
  * Updated swap sorting and display to prioritize claim and refund actions.
  * Improved backup verification and rescue file handling for better reliability.

* **Bug Fixes**
  * Corrected navigation and UI labels to consistently use "rescue" instead of "refund".
  * Fixed handling and validation of swap claim and refund actions.

* **Tests**
  * Added and updated end-to-end tests to cover rescue and claim flows, including new scenarios for BTC and LBTC swaps.
  * Improved test selectors for more robust UI testing.

* **Chores**
  * Updated internal and test utilities to align with new rescue and claim logic.
  * Maintained backward compatibility by redirecting legacy refund URLs to new rescue routes.

PR #991, opened 2025-08-12
fix: Restore previous `SwapChecker` behavior

kiocos:
A regression was introduced in #973, replacing an existing condition in `SwapChecker` with `isSwapClaimable`, but `isSwapClaimable` did not include all swap status which were prevoiusly defined in `SwapChecker`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded swap claim eligibility so reverse swaps become claimable after invoice settlement and chain swaps after transaction claim, reducing stuck swaps and incorrect "not claimable" states.

* **Reliability**
  * Improved blockchain request timeout handling to reduce hangs and speed up responses.

* **Tests**
  * Added tests covering claim eligibility across swap types and statuses.

### DIFF
--- a/src/utils/blockchain.ts
+++ b/src/utils/blockchain.ts
@@ -10,6 +10,10 @@ export type UTXO = {
     vout: number;
 };
 
+const requestTimeout = () => ({
+    signal: AbortSignal.timeout(10_000),
+});
+
 export const fetchUTXOsWithFailover = async (
     asset: string,
     address: string,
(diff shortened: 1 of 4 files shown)


src/utils/blockchain.ts
```diff
@@ -10,14 +10,19 @@ export type UTXO = {
     vout: number;
 };
 
+const requestTimeout = { signal: AbortSignal.timeout(10_000) };
```
michael1011, line 13: As rabbit said, the AbortSignal needs to be recreated for each call

src/utils/rescue.ts
```diff
@@ -41,17 +41,50 @@ export enum RescueAction {
...
+            const pendingStatuses = [
+                swapStatusPending.TransactionServerConfirmed,
+                swapStatusPending.TransactionServerMempool,
+            ];
+
+            if (includeSuccess) {
+                return [
+                    swapStatusSuccess.TransactionClaimed,
+                    ...pendingStatuses,
+                ].includes(status);
```
michael1011, line 79: Same here

Issue #799, opened 2025-01-13
Allow for refunds independent of backend


kilrau:
Dependency: https://github.com/BoltzExchange/boltz-web-app/issues/521, once we have alternative endpoints available and backend is not online/cooperating web-app should fall back to these endpoint to find a lockup transaction, also should show a nice message to the user about timeout block height and ETA for the script path refund.

kilrau:
Related: https://github.com/BoltzExchange/boltz-web-app/pull/828

PR #995, opened 2025-08-13
feat: Backend independent uncooperative refund for timed out swaps

kiocos:


src/utils/rescue.ts
```diff
@@ -6,16 +6,25 @@ import type { ECPairInterface } from "ecpair";
...
 import log from "loglevel";
 
-import { LBTC } from "../consts/Assets";
+import {
+    type AssetType,
+    type BTC,
+    LBTC,
+    LN,
+    type RBTC,
+} from "../consts/Assets";
```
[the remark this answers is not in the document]
kiocos: they are really just used as types tho, so this is fine

src/pages/Pay.tsx
```diff
@@ -123,11 +142,14 @@ const Pay = () => {
         string | undefined
     >(undefined);
 
+    const isWaitingForRefund = () =>
+        swap() &&
+        (swap().type === SwapType.Chain ||
+            swap().type === SwapType.Submarine) &&
+        swapStatus() === swapStatusFailed.TransactionRefunded;
```
michael1011, line 149: Why `TransactionRefunded`? It has nothing to do with `waitingForRefund`

src/utils/blockchain.ts
```diff
@@ -1,6 +1,8 @@
 import log from "loglevel";
 
 import { chooseUrl, config } from "../config";
+import { BTC } from "../consts/Assets";
+import { LBTC } from "../consts/Assets";
```
michael1011, line 5:
```suggestion
import { BTC, LBTC } from "../consts/Assets";
```

src/utils/blockchain.ts
```diff
@@ -134,6 +136,17 @@ const getRawTransaction = async (asset: string, txid: string) => {
...
+export const getBlockTipHeight = async (asset: string) => {
+    return await fetchBlockExplorer<string>(asset, "/blocks/tip/height");
+};
+
+export const getFeeEstimationsExternal = async (asset: string) => {
+    return await fetchBlockExplorer<Record<string, number>>(
+        asset,
+        "/fee-estimates",
+    );
+};
```
michael1011, line 148: This does not exist on the mempool.space API
image

Which would leave blockstream.info as single point of failure here

Also, the esplora fee estimations are super weird, on fee spikes. How about trying or API first, and then falling back to mempool.space and if that doesn't work either, trying Esplora?

src/utils/blockchain.ts
```diff
@@ -154,14 +167,27 @@ export const getSwapUTXOs = async (swap: ChainSwap | SubmarineSwap) => {
...
+                    swap.type === SwapType.Chain
+                        ? (swap as ChainSwap).lockupDetails.timeoutBlockHeight
+                        : (swap as SubmarineSwap).timeoutBlockHeight,
+            };
+        }
+
+        return { hex: rawTx };
+    });
+
+    return transactions;
```
michael1011, line 192:
```suggestion
    return rawTxs.map((rawTx) => {
        if ([BTC, LBTC].includes(swap.assetSend)) {
            return {
                hex: rawTx,
                // Important to know if the swap has timed out or not
                timeoutBlockHeight:
                    swap.type === SwapType.Chain
                        ? (swap as ChainSwap).lockupDetails.timeoutBlockHeight
                        : (swap as SubmarineSwap).timeoutBlockHeight,
            };
        }

        return { hex: rawTx };
    });
```

src/utils/fees.ts
```diff
@@ -0,0 +1,27 @@
...
+import { getFeeEstimations } from "./boltzClient";
+
+export const getFeeEstimationsFailover = async (asset: string) => {
+    try {
+        const feeEstimations = await getFeeEstimations();
+
+        return feeEstimations[asset];
+    } catch (e) {
+        log.warn(
+            `failed to get fee estimations via Boltz client for ${asset}: ${e}`,
```
michael1011, line 13:
```suggestion
            `failed to get fee estimations via Boltz API for ${asset}: ${e}`,
```

src/utils/helper.ts
```diff
@@ -109,12 +114,16 @@ export const fetcher = async <T = unknown>(
...
             },
+            signal: controller.signal,
             body: JSON.stringify(params),
         };
     }
 
     const apiUrl = getApiUrl() + url;
     const response = await fetch(apiUrl, options || opts);
+
+    clearTimeout(requestTimeout);
```
michael1011, line 125: These *always* have to go in `finally` block

src/utils/rescue.ts
```diff
@@ -370,10 +389,82 @@ export const getRefundableUTXOs = async (currentSwap: SomeSwap) => {
...
 };
 
+const getCurrentBlockHeight = async (swaps: SomeSwap[]) => {
+    try {
+        const assetSet = new Set<AssetType>(
+            swaps.map((swap) => swap.assetSend as AssetType),
+        );
+
+        type RefundableAsset = typeof BTC | typeof LBTC | typeof RBTC;
+        const assets: RefundableAsset[] = Array.from(assetSet).filter(
```
michael1011, line 399:
```suggestion
        const assets: RefundableAsset[] = Array.from(swaps.map((swap) => swap.assetSend as AssetType)).filter(
```

src/utils/rescue.ts
```diff
@@ -370,10 +389,82 @@ export const getRefundableUTXOs = async (currentSwap: SomeSwap) => {
...
+                return;
+            }
+
+            const height = Number(res.value);
+            if (!Number.isFinite(height)) {
+                log.warn(
+                    `invalid block tip height for asset ${asset}: ${res.value}`,
+                );
+                return;
+            }
```
michael1011, line 422: This should be done in `getBlockTipHeight`

michael1011 (changes requested):

src/pages/Pay.tsx
```diff
@@ -65,13 +69,25 @@ const Pay = () => {
...
         }
+
+        setSwap(currentSwap);
+
+        // If the swap timeoutBlockHeight was reached, we don't rely on backend for the swap status
+        if (timedOutRefund) {
+            log.info(
+                `refundable swap ${currentSwap.id} timed out, uncooperative refund is possible`,
+            );
+            setSwapStatus(swapStatusFailed.TransactionRefunded);
```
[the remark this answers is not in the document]
kiocos: @michael1011 We used to map `TransactionRefunded` to `swap.waitingForRefund` and I did some confusion with that, but you're right, it doesn't make sense.

I decided to add the status `waitingForRefund` to the `swapStatusFailed` object and use that. LMK if you agree with this or if `swapStatusFailed` should be exclusively for statuses returned from the backend.

kiocos:
@michael1011 pushed new code from [fixup](https://github.com/BoltzExchange/boltz-web-app/pull/995/commits/a7ba51b342ab7a2a05ee3ceb01779af33f37758c) onwards

kilrau:
Missing https://github.com/BoltzExchange/boltz-web-app/issues/799 closing statement in pr description

kilrau (changes requested):
I am on regtest with boltz-backend container down. So while when I enter via Rescue -> Refund web app correctly detects my swap as refundable:
Image

But when simply refreshing above correct state ( https://localhost:5173/swap/6wG7tkBcNPGR) web app gets thrown off by the failing websocket and never checks/loads the uncooperative refund screen and is stuck in this loop:
https://github.com/user-attachments/assets/d581c4dd-8d70-47bd-8d0f-f52964f69d17

Same result if I enter the swap via the history menu.

While I understand why you only check when entering via this path and not on `/swap`, I think it makes sense to check if swap timeout was reached to bring up the uncooperative refund screen when boltz api/ws is down. Wdyt?

Also: if Boltz Backend is up and pretends to cooperate in refund but doesn't then web app will be stuck the same way I guess?

kilrau (changes requested):
Well I tried uncooperative refund with a couple different swaps, all failing with `NaN can't be converted to BigInt because it isn't an integer`.

Image

Image

kilrau (approved):
dc650d35dcee924d126445cb85bc3f7db271fd27 working nicely:

Image

33b997e87768d79666fe978de3c12e0e6f2b9f5e too:

Image

src/utils/blockchain.ts
```diff
@@ -134,6 +154,120 @@ const getRawTransaction = async (asset: string, txid: string) => {
...
+    ) => {
+        throw new Error(
+            `unexpected fee estimation response for ${asset}: ${JSON.stringify(
+                feeEstimations,
+            )}`,
+        );
+    };
+
+    // we don't have mempool on regtest
+    if (config.network !== "regtest") {
```
michael1011, line 235: Why hardcode that. Just don't try to use mempool, when we don't have a mempool instance configured

src/utils/blockchain.ts
```diff
@@ -238,6 +255,10 @@ export const getExplorerFeeEstimations = async (asset: string) => {
 
         const expectedBlocksToConfirm = "3";
 
+        if (typeof feeEstimations[expectedBlocksToConfirm] !== "number") {
+            handleUnexpectedResponse(feeEstimations);
```
michael1011, line 259: Throwing inside of the function that fetches the estimation means you don't have to remember and repeat this logic everywhere you use it

michael1011 (changes requested):

kilrau:
So I just tested and can confirm that "swaps showing as refundable" behavior is correct and in line with master now: no more false positives that show as refundable. [The fix](https://github.com/BoltzExchange/boltz-web-app/pull/995/commits/a705289e7bf6cd9b5a9d4820c87f56f5291e78ec) looks legit to me, any swap that had a lockuptx was marked as refundable before, but only swap addresses with utxos should. But @michael1011 has to vouch for it.

michael1011 (approved):

PR #999, opened 2025-08-22
Mnemonic input improvements

### LINKED ISSUE #979: Invalid rescue key when pasting mnemonic with new liner at the end
When testing the WebApp, I accidentally copied a new liner from the seed and pasted it into the rescue key input
Image

Image

It shows a "Invalid rescue key" error, but it doesn't display the pasted new liner, which makes it hard to guess it is there.

In the above example, pressing backspace just deletes the `e` from `cube`. Adding the `e` again makes the validation pass.

WebApp should properly trim newliners.

kiocos:
Closes #979


src/components/MnemonicInput.tsx
```diff
@@ -74,7 +74,7 @@ const MnemonicInput = (props: { onSubmit: (mnemonic: string) => void }) => {
 
     const handlePaste = (e: ClipboardEvent) => {
         e.preventDefault();
-        const pastedText = e.clipboardData.getData("text/plain");
+        const pastedText = e.clipboardData.getData("text/plain").trim();
         const words = pastedText.split(" ");
```
michael1011, line 78: We should also split on `\n`

src/components/MnemonicInput.tsx
```diff
@@ -74,7 +74,7 @@ const MnemonicInput = (props: { onSubmit: (mnemonic: string) => void }) => {
 
     const handlePaste = (e: ClipboardEvent) => {
         e.preventDefault();
-        const pastedText = e.clipboardData.getData("text/plain");
+        const pastedText = e.clipboardData.getData("text/plain").trim();
```
michael1011, line 77: This needs a test case to avoid regressions

kilrau (approved):
Works. Fine to merge after addressing Michael's comments.

PS: Technically we could trim all non-chars, e.g. when pasting the mnemonic from the rescue key file sth like this might happen
Image

Also, if the pasted content ends with space+something the entire thing breaks and all content is just pasted in the first input field. Desirable behavior would be to populate as-is and have all invalid content in the 12th input field instead.
Image

But both pretty nit picky and nothing you should spend time on. Just wanted to give you some proof I actually tested this :p

PR #1006, opened 2025-08-25
fix: min-relay fee not met

### LINKED ISSUE #985: Uncooperative liquid claims fail with "min relay fee not met"
As discussed, we want to add two sats buffer, like we do for unconfidential liquid already and thus overpay 2 sats in the cooperative scenario. Only way to have exact receive amounts AND make uncoooperative claims work reliably.

And we really **need tests** for this!!

For context here our discussion:
> @kiocos 
> Regarding insufficient relay fee:
> I added the throw on the cooperative claim path for BTC->L-BTC swap and the problem always happen, even on the regular swap flow (claiming with existing data, without restoring). 
> Backend locks its transaction paying 27 sats of fee.
>
> In the happy scenario: WebApp is able to cooperatively claim the locked transaction (using 20 sats for fee).
> In the bad scenario: WebApp fails to uncooperatively claim the locked transaction (using 21 sats for fee).
> The fee is calculated with: const feeBudget = Math.floor(inputSum - swap.receiveAmount);.
> I see that swap.receiveAmount has the correct value, thus ig the problem should be on inputSum. 
> 
> Could the problem be that backend is locking its tx with a fee higher than it should?
> Had it locked it with 25 sats, it would leave 2 additional sats for inputSum, which would result in  a feeBudget of 23 for the uncooperative claim. 
> 

> @michael1011 
> The backend does exactly what you ask it to
> 
> You get exactly the amount you agreed on in the api
> 
> @kilrau  so do you want exact receive amounts OR script path enforcement being possible
> 
> YOU CANNOT HAVE BOTH 
> 
> Or you add two sats buffer, like we do for unconfidential liquid already
> 
> And overpay fees in the cooperative scenario 

> @kilrau 
> let's go for slightly overpaying fees in the cooperative scenario

kiocos:
Closes #985

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted miner fee handling for BTC↔LBTC routes; users may see a small increase in estimated/claimed miner fees when fetching pairs and during swaps (applies to both regular and non-regular pair fetches).

* **Tests**
  * Updated expected receive amount and zero-amount BTC/L-BTC expectation to reflect the revised fee calculations.

### DIFF
--- a/src/context/Global.tsx
+++ b/src/context/Global.tsx
@@ -14,6 +14,7 @@ import {
 import type { Accessor, JSX, Setter } from "solid-js";
 
 import { config } from "../config";
+import { LBTC } from "../consts/Assets";
 import { Denomination, UrlParam } from "../consts/Enums";
 import { referralIdKey } from "../consts/LocalStorage";
 import { swapStatusFinal } from "../consts/SwapStatus";
@@ -34,6 +35,7 @@ import { getUrlParam, isEmbed, resetUrlParam } from "../utils/urlParams";
 import { checkWasmSupported } from "../utils/wasmSupport";
 import { detectWebLNProvider } from "../utils/webln";
 
+export const liquidUncooperativeExtra = 3;
 const proReferral = "pro";
 
 export type deriveKeyFn = (index: number) => ECPairInterface;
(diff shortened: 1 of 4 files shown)


src/context/Global.tsx
```diff
@@ -292,9 +294,19 @@ const GlobalProvider = (props: { children: JSX.Element }) => {
         void audio.play();
     };
 
+    const addUncooperativeExtra = (pairs: Pairs) => {
+        pairs.chain[BTC][LBTC].fees.minerFees.user.claim +=
+            liquidUncooperativeExtra;
+        pairs.reverse[BTC][LBTC].fees.minerFees.claim +=
+            liquidUncooperativeExtra;
```
michael1011, line 301: Iterate over all pairs and add the extra when our receiving side is Liquid

PR #1027, opened 2025-09-15
fix: Trezor's tx value's NaN

kiocos:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed transaction signing issues with Trezor hardware wallets by properly hex-prefixing the transaction value. This improves compatibility with Trezor Connect and prevents “invalid value” errors during signing.
  * No changes to gas, nonce, or final on-chain amounts; this is a formatting correction to ensure successful signing.

### DIFF
--- a/src/utils/hardware/TrezorSigner.ts
+++ b/src/utils/hardware/TrezorSigner.ts
@@ -114,7 +114,7 @@ class TrezorSigner implements EIP1193Provider, HardwareSigner {
                     nonce: nonce.toString(16),
                     chainId: Number(network.chainId),
                     gasPrice: feeData.gasPrice.toString(16),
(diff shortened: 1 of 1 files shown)


kilrau:
Nice find! If user confirmed that this fixed it: As discussed yesterday, this may brick old Trezor libraries and can't be merged like this. It'd need to check for the 0x prefix and only add it if it's missing.

michael1011:
> this may brick old Trezor libraries

Why?

michael1011:
> This was working before without the 0x prefix so I'm afraid it'll brick

Yeah, we are not playing guessing games here. It's working fine with a Trezor Model One and Safe with the latest firmware

PR #1033, opened 2025-10-02
feat: Show lnurl/lnaddress min/max amount errors

kiocos:


src/i18n/i18n.ts
```diff
@@ -310,6 +310,14 @@ const dict = {
...
         get_refundable_error:
             "Failed to load UTXO data. Refresh to try again or check your internet connection if the problem persists.",
+        min_amount_lnurl:
+            "Minimum amount for this lnUrl is {{ amount }} {{ denomination }}",
+        max_amount_lnurl:
+            "Maximum amount for this lnUrl is {{ amount }} {{ denomination }}",
+        min_amount_lnaddress:
+            "Minimum amount for this lnAddress is {{ amount }} {{ denomination }}",
+        max_amount_lnaddress:
+            "Maximum amount for this lnAddress is {{ amount }} {{ denomination }}",
```
michael1011, line 320: It's "LNURL" and "Lightning Address". But I really don't see a point in distinguishing between them in error messages. What i meant in https://github.com/BoltzExchange/boltz-web-app/pull/1033#discussion_r2402106893 was something like `Minimum amount for the destination is {{ amount }} {{ denomination }}`

michael1011 (changes requested):

michael1011 (approved):
Test should be fixed and needs translations; can be merged after

kilrau (changes requested):
Works, but denomination here needs to be removed - it's always Lightning
image

PR #1034, opened 2025-10-07
chore: log created swap data

kiocos:


kiocos:
> Claim address works, but lightning invoice for submarine swaps should be logged the same way. Currently it isn't. Didnt' test Rootstock.

Fixed. Here's Rootstock:
image

Here's LN:
image

PR #1035, opened 2025-10-07
feat: Show receiving address/invoice on swap send screen

kiocos:
image

image


kiocos:
@kilrau needs translations

kilrau (changes requested):
As-is impossible to properly read longer destinations like invoices, also it disappears when moving mouse off. I really want to be able to mark and copy the text. Try something similar to the MRH optimized swap hint (i). Make it stay open, let me mark and copy, check that long strings like invoices are completely visible and formatted properly.

Image

michael1011 (commented):
I like it. Regarding it being "confusing", i guess we could hide it by default and only show it in an "advanced mode" that can be enabled in the tooltip? But maybe we can come up with something nicer. Tbh, I am fine with this though

PR #1050, opened 2025-10-22
fix #1047 - Display all characters of deposit address

### LINKED ISSUE #1047: Display all characters of deposit address
Display all characters of the deposit address, separated in groups of 4 or 5 characters, like `bcrt1 qrhg8 z3ccu 8vmnz 7xvwx 8t92m ykw6r u64k9 6e4v`.

dannym-arx:
Closes #1047 

would this implementation satisfy the requirements?

Image

Image

(copy buttons still work as expected, copying without spaces naturally)

in the input field, the address only formats upon pasting (or typing) a valid address, if the address is invalid it won't format, to signal the user something's wrong (besides the red border) (thought that would be a good ux, but if not it's a quick change)

## Summary by CodeRabbit

* **New Features**
  * Address values are now formatted into readable groups of characters; invoice values retain their previous shortened display.

* **Style**
  * Improved alignment and text-wrapping in copy components for better visual presentation.

* **Tests**
  * Added tests for address formatting and for copy behavior verifying the raw address is copied to the clipboard.

### DIFF
--- a/src/components/CopyBox.tsx
+++ b/src/components/CopyBox.tsx
@@ -1,6 +1,7 @@
 import { BiRegularCopy } from "solid-icons/bi";
 import { IoCheckmark } from "solid-icons/io";
 import { Show, createSignal } from "solid-js";
+import { isInvoice } from "src/utils/invoice";
 
 import { copyIconTimeout } from "../consts/CopyContent";
 import {
@@ -33,7 +34,7 @@ const CopyBox = (props: { value: string }) => {
                 fallback={<BiRegularCopy size={23} data-testid="copy-icon" />}>
                 <IoCheckmark size={23} data-testid="checkmark-icon" />
             </Show>
-            {props.value.startsWith("ln")
+            {isInvoice(props.value)
                 ? cropString(props.value, baseStrLength, maxStrLength)
                 : formatAddress(props.value)}
         </p>


src/utils/helper.ts
```diff
@@ -59,6 +59,16 @@ export const cropString = (str: string, maxLen = 40, subStrSize = 19) => {
     );
 };
 
+export const formatAddress = (address?: string | null | undefined): string => {
```
kiocos:
```suggestion
export const formatAddress = (address?: string | null): string => {
```

src/components/AddressInput.tsx
```diff
@@ -109,7 +154,7 @@ const AddressInput = () => {
             name="onchainAddress"
             autocomplete="off"
             placeholder={t("onchain_address", { asset: assetReceive() })}
-            value={onchainAddress()}
+            value={formatAddress(onchainAddress())}
```
kiocos: Not sure if we really want to do this format here. We already have the `Destination` tooltip at the payment page which serves to help the user double-check the input.

@kilrau ?

If we wish to do so, I've noticed two nits to fix here:
1. The keyboard cursor doesn't work properly when using the left and right arrows inside this input (behaves weird at the spaces);
2. `Delete` button also doesn't work when we press it at one of the blank spaces;

Otherwise, we can revert the changes from `AddressInput`

kiocos (changes requested):
Thank you for your PR!

CI is currently breaking due to linting errors needing a fix. Let's also sign all the commits! 💪

Not sure if we want to apply the format on `AddressInput`, see https://github.com/BoltzExchange/boltz-web-app/pull/1050#discussion_r2453576489

Let's also apply these changes to chain addresses only, we should keep the previous behavior for invoices (no need to display the full invoice).

kiocos (changes requested):
Looking good! :muscle: 
Just one last nit, IMO can be merged after addressed.
<details>
<summary>screenshots</summary>
Image

image

Image
</details>

PR #1065, opened 2025-11-12
fix: RescueExternal params

kiocos:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redirect handling in refund/rescue flows so redirects respect current route parameters and preserve query/search strings across external and internal refund scenarios.
* **Tests**
  * Expanded end-to-end coverage to verify rescue-key query behavior across external rescue routes, ensuring proper UI behavior for different external types.


michael1011 (commented):
Let's add a E2E test to avoid regressions

michael1011 (approved):

PR #1067, opened 2025-11-14
feat: Show fiat rate (BTC/USD)

### LINKED ISSUE #372: Add subtle USD values to swap box
https://twitter.com/Boltzhq/status/1736338124941222177
image which translates to "I have a proposal that you add a conversion function, for example, if I need to transfer 500 US dollars of BTC, then your system will calculate the corresponding SATS according to the conversion."

here USD amount comparison would help to make sure the btc to sat denomination was done correctly:
image

kiocos:
Closes #372 

## Summary by CodeRabbit

* **New Features**
  * Display Bitcoin amounts in USD fiat equivalents throughout the application
  * Toggle fiat rate display on/off from the settings menu
  * Full multi-language support for fiat features
  * Automatic Bitcoin price fetching with failover when providers are unavailable

* **Tests**
  * Added comprehensive tests for fiat conversion and price fetching

### DIFF
--- a/src/components/PayInvoice.tsx
+++ b/src/components/PayInvoice.tsx
@@ -1,6 +1,6 @@
 import { BigNumber } from "bignumber.js";
 import log from "loglevel";
-import { Show, createEffect } from "solid-js";
+import { Show } from "solid-js";
 
 import CopyButton from "../components/CopyButton";
 import QrCode from "../components/QrCode";
@@ -13,8 +13,7 @@ import { enableWebln } from "../utils/webln";
 import CopyBox from "./CopyBox";
 
 const PayInvoice = (props: { sendAmount: number; invoice: string }) => {
-    const { t, denomination, separator, webln, fetchBtcPrice } =
-        useGlobalContext();
+    const { t, denomination, separator, webln } = useGlobalContext();
 
     const payWeblnInvoice = async (pr: string) => {
         await enableWebln(async () => {
@@ -22,10 +21,6 @@ const PayInvoice = (props: { sendAmount: number; invoice: string }) => {
             log.debug("webln payment result:", result);
         });
     };
-
-    createEffect(() => {
-        void fetchBtcPrice();
-    });
 
     return (
         <div>
--- a/src/components/PayOnchain.tsx
+++ b/src/components/PayOnchain.tsx
@@ -1,5 +1,5 @@
 import { BigNumber } from "bignumber.js";
-import { Show, createEffect, createMemo, createResource } from "solid-js";
+import { Show, createMemo, createResource } from "solid-js";
 
 import CopyButton from "../components/CopyButton";
 import QrCode from "../components/QrCode";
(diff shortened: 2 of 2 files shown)


src/configs/base.ts
```diff
@@ -75,6 +75,11 @@ const defaults = {
     dnsOverHttps: "https://1.1.1.1/dns-query",
     chatwootUrl: "https://support.boltz.exchange",
     preimageValidation: "https://validate-payment.com",
+    rateProviders: {
+        Yadio: "https://api.yadio.io/exrates/btc",
+        Kraken: "https://api.kraken.com/0/public/Ticker",
```
michael1011: Why kraken?

src/context/Global.tsx
```diff
@@ -470,6 +472,14 @@ const GlobalProvider = (props: { children: JSX.Element }) => {
         },
     );
 
+    const [showFiatAmount, setShowFiatAmount] = makePersisted(
+        // eslint-disable-next-line solid/reactivity
+        createSignal<boolean>(true),
```
michael1011: Let's disable that by default

src/i18n/i18n.ts
```diff
@@ -196,6 +196,10 @@ const dict = {
         denomination_tooltip: "Choose your preferred denomination: BTC or sats",
         decimal_tooltip:
             "Choose your preferred decimal separator: dot or comma",
+        show_fiat_rate: "Show Fiat Rate",
+        show_fiat_rate_tooltip:
+            "Enable or disable the display of fiat rate (USD) for swap amounts",
```
michael1011:
```suggestion
            "Show USD amounts",
```

src/utils/fiat.ts
```diff
@@ -0,0 +1,76 @@
...
+    } catch (e) {
+        throw new Error("failed to get BTC price from Mempool: ", e);
+    }
+};
+
+export const getBtcPriceFailover = async (
+    currency: Currency = Currency.USD,
+) => {
+    for (const getBtcPrice of [
+        getBtcPriceYadio,
```
michael1011: I am not comfortable using that random site by default as first choice

michael1011 (changes requested):

kiocos:
> LGTM
> 
> But did you now toggle it off by default on purpose? If so, why?

Yes. See https://github.com/BoltzExchange/boltz-web-app/pull/1067#discussion_r2529839304

src/context/Global.tsx
kilrau: Why? I really think it's hyper subtle and noone will ever manually enable this as it's not discoverable.

kilrau:
> > LGTM
> > But did you now toggle it off by default on purpose? If so, why?
> 
> Yes. See [#1067 (comment)](https://github.com/BoltzExchange/boltz-web-app/pull/1067#discussion_r2529839304)

I really think it should be on by default. It's useful without thinking much about it, let alone go to settings and check for the option. It'll have close to 0 effect if we leave it off by default. Check https://app.uniswap.org, subtle and on by default.

kilrau (approved):
tACK

michael1011 (approved):

PR #1068, opened 2025-11-19
fix: Support RSK refund via External Rescue

kiocos:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness in refund/claim flows by safely handling missing claim-related data so refunds don't fail when details are absent.
  * Expanded refund eligibility logic to include additional swap types that were previously excluded, ensuring more swaps are considered refundable.


michael1011 (approved):
tACK

PR #1073, opened 2025-11-24
refactor: BIP21 parameter extraction

### LINKED ISSUE #1071: Refactor BIP21 parameter extraction
@michael1011 I'd rather not change this in this PR. `compat.ts` and `invoice.ts` have a bunch of functions which don't guard against `null` and just assume the value passed is a `string`.

Returning `null` here could cause me to pass `null` to a bunch of those, which could cause regressions if I miss some.

For example, it would break in cases like this:
```tsx
export const isInvoice = (data: string) => {
    const prefix = bolt11Prefixes[config.network];
    const startsWithPrefix = data.toLowerCase().startsWith(prefix);
```

_Originally posted by @kiocos in https://github.com/BoltzExchange/boltz-web-app/pull/1066#discussion_r2548060628_

kiocos:
Closes #1071 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input validation with defensive type checks to prevent errors from non-string or malformed inputs.
  * Removed duplicated BIP21 handling so BIP21 amounts now update receive/send values once and earlier in the input flow.

* **Tests**
  * Updated unit test expectation to return null for certain BIP21 invoice extraction cases.
  * Added end-to-end tests validating BIP21, Lightning, and BOLT12 URI parsing and related UI behavior.

### DIFF
--- a/e2e/create.spec.ts
+++ b/e2e/create.spec.ts
@@ -160,7 +160,7 @@ test.describe("BIP21 URIs", () => {
         await page.getByTestId("select-LN").click();
 
         const bolt12Offer = await getBolt12Offer();
-        const bip21Uri = `lno:${bolt12Offer}`;
+        const bip21Uri = `lightning:${bolt12Offer}`;
 
         const invoiceInput = page.getByTestId("invoice");
         await invoiceInput.fill(bip21Uri);


michael1011 (commented):
tACK

But needs tests

kilrau (approved):
Did some "manual fuzz tests". Looking **good!**

kilrau:
Currently not ideal behavior: invalid invoices or addresses break behavior it should always switch to the currency and then cleanly error. Examples:

1. Should choose Lightning, paste invoice string and error: `BITCOIN:BC1QYLH3U67J673H6Y6ALV70M0PL2YZ53TZHVXGG7U?amount=0.0005&lightning=lnbc5000u1p3pj257pp5test`
2. Same, should trigger Lightning and error: `lightning:lnbc10u1p3pj257pp5test`
3. Should trigger Bitcoin and throw error: bitcoin:BC1QTEST?amount=0.001

PR #1084, opened 2025-12-01
feat: add zero-conf toggle

### LINKED ISSUE #346: 0-conf toggle & notifications

kiocos:
Closes #346 

## Summary by CodeRabbit

* **New Features**
  * Zero-Conf toggle in Security settings to enable automatic claiming of eligible swaps.
  * Claiming API added to the payment context to coordinate cooperative/server-assisted claims.

* **Localization**
  * Added Zero-Conf label and tooltip translations; removed many legacy swap-related keys.

* **Behavior Changes**
  * Rescue list and claim eligibility respect Zero-Conf state.

* **Refactor**
  * Claim logic moved from UI checker into the payment/context layer.

* **Tests**
  * Unit and E2E tests updated and new E2E scenarios added for Zero-Conf auto-claim.

### DIFF
--- a/src/context/Pay.tsx
+++ b/src/context/Pay.tsx
@@ -163,7 +163,7 @@ const PayProvider = (props: { children: JSX.Element }) => {
 
         if (
             (zeroConf() &&
-                data.status === swapStatusPending.TransactionMempool) ||
+                data.status === swapStatusPending.TransactionMempool) || // necessary for the autoclaim when zeroConf is toggled with a pending swap
             data.status === swapStatusSuccess.InvoiceSettled
         ) {
             data.transaction = await getReverseTransaction(currentSwap.id);
@@ -171,7 +171,7 @@ const PayProvider = (props: { children: JSX.Element }) => {
             currentSwap.type === SwapType.Chain &&
             (data.status === swapStatusSuccess.TransactionClaimed ||
                 (zeroConf() &&
-                    data.status === swapStatusPending.TransactionServerMempool))
+                    data.status === swapStatusPending.TransactionServerMempool)) // necessary for the autoclaim when zeroConf is toggled with a pending swap
         ) {
             data.transaction = (
                 await getChainSwapTransactions(currentSwap.id)


src/utils/rescue.ts
```diff
@@ -95,7 +97,7 @@ export const isSwapClaimable = ({
         case SwapType.Reverse: {
             const statuses = [
                 swapStatusPending.TransactionConfirmed,
-                swapStatusPending.TransactionMempool,
+                ...(zeroConf ? [swapStatusPending.TransactionMempool] : []),
```
michael1011: Why not just push to it?

src/utils/rescue.ts
```diff
@@ -107,7 +109,9 @@ export const isSwapClaimable = ({
         case SwapType.Chain: {
             const statuses = [
                 swapStatusPending.TransactionServerConfirmed,
-                swapStatusPending.TransactionServerMempool,
+                ...(zeroConf
+                    ? [swapStatusPending.TransactionServerMempool]
+                    : []),
```
michael1011: Same here

kiocos:
@michael1011 in order to make the claim functionality reactive to the zeroConf toggle, I moved `claimSwap` from `SwapChecker` to `context/Pay.tsx` to share the claiming logic.

I've added `const claimingSwaps = new Set<string>();` because claiming is async and the WebApp receives a WebSocket message regarding the swap status update, which triggers another claim.

The duplicated claim request was being performed even before the code below (from the first claim) finished executing:
```
claimedSwap.claimTx = res.claimTx;
await setSwapStorage(claimedSwap);
```
Having the `claimTx` set would have prevented a duplicated request, if it was available earlier.
The `claimingSwaps` was the solution I found to fix this.

src/pages/RescueExternal.tsx
```diff
@@ -148,7 +148,7 @@ export const RefundBtcLike = () => {
         currentSwaps,
         async (swaps: SomeSwap[]) => {
             setLoading(true);
-            return await createRescueList(swaps).finally(() =>
+            return await createRescueList(swaps, zeroConf()).finally(() =>
```
michael1011: I would also also accept 0-conf in the external rescues

src/context/Pay.tsx
```diff
@@ -45,6 +100,198 @@ const PayProvider = (props: { children: JSX.Element }) => {
...
+            (zeroConf() &&
+                data.status === swapStatusPending.TransactionMempool) ||
+            data.status === swapStatusSuccess.InvoiceSettled
+        ) {
+            data.transaction = await getReverseTransaction(currentSwap.id);
+        } else if (
+            currentSwap.type === SwapType.Chain &&
+            (data.status === swapStatusSuccess.TransactionClaimed ||
+                (zeroConf() &&
+                    data.status === swapStatusPending.TransactionServerMempool))
```
michael1011: Same here

kiocos: If I turn zero-conf on in this situation, it wouldn't autoclaim with the previous condition because its status is `TransactionServerMempool` and not `TransactionClaimed`. The same for the other thread.
image

michael1011: Comment in the code please

PR #1089, opened 2025-12-03
feat: show originally entered ln destination on funding screen instead of resolved invoice

### LINKED ISSUE #1086: Show originally entered ln destination on funding screen instead of resolved invoice
Recently had a support case with a user being confused because when creating a swap, he used a lightning address:

Image

But then on the swap funding screen an invoice is shown as destination, which "is different". And he's right.

Image

We just agreed to show whatever was used at swap creation BIP353/LNURL/Lightning address and not the final resolved invoice on the second screen as "Destination" to avoid confusion of novice users. Advanced users know anyways.

kilrau:
Closes #1086. Added to Chain swaps for MRH resolution and to Reverse Swap just because it doesn't hurt and to be future proof, who knows what we'll come up with.

image

## Summary by CodeRabbit

* **New Features**
  * Preserve and prefer the original payment destination across submarine, reverse, and chain swap flows to improve payment traceability and reduce misrouting; original destination is cleared during swap cleanup/reset.

* **Tests**
  * Added tests validating destination selection and fallback behavior across swap and invoice scenarios to ensure reliability.

### DIFF
--- a/src/utils/helper.ts
+++ b/src/utils/helper.ts
@@ -209,7 +209,13 @@ export const getDestinationAddress = (swap: SomeSwap) => {
     }
 
     if (swap.type === SwapType.Submarine) {
-        return (swap as SubmarineSwap).invoice;
+        const submarineSwap = swap as SubmarineSwap;
+        return submarineSwap.originalDestination || submarineSwap.invoice;
+    }
+
+    if (swap.type === SwapType.Reverse || swap.type === SwapType.Chain) {
+        const chainSwap = swap as ReverseSwap | ChainSwap;
+        return chainSwap.originalDestination || chainSwap.claimAddress;
     }
 
     return swap.claimAddress;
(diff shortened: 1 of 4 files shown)


src/utils/helper.ts
```diff
@@ -209,7 +209,18 @@ export const getDestinationAddress = (swap: SomeSwap) => {
...
+    }
+
+    if (swap.type === SwapType.Chain) {
+        const chainSwap = swap as ChainSwap;
+        return chainSwap.originalDestination || chainSwap.claimAddress;
+    }
+
+    if (swap.type === SwapType.Reverse) {
+        const reverseSwap = swap as ReverseSwap;
+        return reverseSwap.originalDestination || reverseSwap.claimAddress;
```
michael1011: These two have an overlap

```suggestion
    if (swap.type === SwapType.Reverse || SwapType.Chain) {
        const chainSwap = swap as ReverseSwap | ChainSwap ;
        return chainSwap.originalDestination || chainSwap.claimAddress;
```

michael1011 (changes requested):

kiocos (changes requested):

kilrau:
Reviews adddressed, ready to re-review

michael1011 (approved):

PR #1111, opened 2025-12-15
feat: detect timeout expiration for uncooperative refund on `/swap` page

kilrau:
It's not great UX to force users to constantly enter via `/rescue` entry to have the block height check trigger instead of having it trigger right on `/swap` on page reload once one reached the timeout ETA page.

I remember us discussing this, so maybe I missed why this is not possible. In any case, please give this a look and either reject with a good reason or help me finish it up (including tests) :pray: @kiocos 

## Summary by CodeRabbit

* **New Features**
  * Refund transaction ID is surfaced in the UI and drives a refunded-state view.

* **Improvements**
  * Automatic refund prompt when swaps time out; improved timeout/ETA handling with graceful fallbacks.
  * Reduced unnecessary UTXO and block-explorer checks in initial/just-claimed/non-refundable cases.
  * Recognizes an invoice-paid status to refine swap progression.
  * Refund flows now propagate refund IDs via component callbacks for clearer UX.

* **Tests**
  * New unit and e2e tests covering timeout, ETA display, automatic refund prompts, and reverse-swap behavior.

### DIFF
--- a/src/pages/Pay.tsx
+++ b/src/pages/Pay.tsx
@@ -195,9 +195,6 @@ const Pay = () => {
             setRefundableUTXOs(utxos);
 
             if (utxos.length > 0) {
-                // if there are remaining UTXOs, we consider we don't have a refundTx yet
-                setSwap({ ...swap(), refundTx: "" });
-
                 if (isRefundableSwapType(swap()) && !timedOutRefundable()) {
                     const timeoutBlockHeight =
                         swap().type === SwapType.Submarine


kiocos:
Noticed the refund screen could be incorrectly shown for Rootstock.  Fixed on [03e87a7](https://github.com/BoltzExchange/boltz-web-app/pull/1111/commits/03e87a7537c4f93872507bb57409438a2bb1b6e3)

src/status/SwapExpired.tsx
```diff
@@ -1,24 +1,35 @@
...
+                <RefundButton
+                    swap={swap as Accessor<SubmarineSwap | ChainSwap>}
+                    setRefundTxId={setRefundTxId}
+                />
+                <hr />
+            </Show>
+            <Show when={refundTxId() !== ""}>
+                <SwapRefunded />
+                <hr />
+            </Show>
```
michael1011: Same thing here for the reverse swaps

src/pages/Pay.tsx
```diff
@@ -92,6 +93,7 @@ const Pay = () => {
     const [copyDestinationActive, setCopyDestinationActive] =
         createSignal(false);
     const [loading, setLoading] = createSignal<boolean>(false);
+    const [justRefunded, setJustRefunded] = createSignal<boolean>(false);
```
michael1011: Why this dirty hack and not just check if we have a refund transaction id?

kiocos: checking if we have a refund transaction id is not enough for the purpose of showing a temporary refund screen (unless we're okay with not keeping the tx id... and now I think we're okay with it, will  test it out)

e2e/rescue/refund.spec.ts
```diff
@@ -237,8 +237,9 @@ const validateRefundTransaction = async (
 test.describe("Refund", () => {
     const refundFileJson = path.join(__dirname, fileName);
 
-    test.beforeEach(async () => {
+    test.beforeEach(async ({ page }) => {
         await generateLiquidBlock();
+        await page.route("**/utxo", (route) => route.continue()); // disabling HTTP caching for UTXOs
```
michael1011: Are you sure this actually does something?

kiocos: Pretty confident it does. From Playwright docs: `Enabling routing disables http cache.`
https://playwright.dev/docs/api/class-browsercontext

It doesn't mention I need to do anything else besides enabling this. Had flaky tests failing constantly prior to this and now they pass reliably.

PR #1142, opened 2026-01-16
fix: browser's navigation buttons on paginated pages

kiocos:
When navigating from any non-paginated page to any paginated page then pressing the browser's back button, the browser would not go back. This PR fixes this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser history handling when pagination resets to the first page due to invalid page parameters, providing better back-button behavior.


kilrau (approved):
tACK

Nice one!

PR #1177, opened 2026-02-24
chore: update to boltz-core v4

### LINKED ISSUE #1133: Update to upcoming `boltz-core`
New Boltz Core release makes WASM optional as only needed for Liquid swaps. 
- We want to allow non-Liquid swaps without requiring "WASM to be enabled"
- Fetching `secp256k1-zkp` only needed for blinding/unblinding of liquid transactions. So we should consider only starting to fetch when a *Liquid* swap is created (consider pending Liquid swaps)

kiocos:
Closes #1133 

## Summary by CodeRabbit

* **Bug Fixes**
  * Verified refunded status message now appears correctly in the UI after a refund is triggered.

* **Refactor**
  * Modernized cryptography, encoding and transaction handling across the app for improved reliability and consistency.

* **Tests**
  * Updated end-to-end and unit tests to cover refund visibility and new byte/hex-based data flows.


src/utils/boltzClient.ts
```diff
@@ -10,6 +9,16 @@ import { broadcastToExplorer } from "./blockchain";
...
 
+export const txToHex = (transaction: TransactionInterface): string =>
+    transaction instanceof LiquidTransaction
+        ? transaction.toHex()
+        : (transaction as BtcTransaction).hex;
+
+export const txToId = (transaction: TransactionInterface): string =>
+    transaction instanceof LiquidTransaction
+        ? transaction.getId()
+        : (transaction as BtcTransaction).id;
```
michael1011: This should go in a utils or compat file. Not here

PR #1182, opened 2026-02-25
feat: resume pending external EVM swaps v2

### LINKED ISSUE #978: Rescue Key: resume pending Rootstock reverse/chain swaps
https://github.com/BoltzExchange/boltz-web-app/pull/973 only covered Liquid/Mainchain. This is a follow-up issue about Rootstock.

kiocos:
Closes #978 

## Summary by CodeRabbit

* **New Features**
  * Unified rescue experience: file/mnemonic upload, RSK scan mode, and a new EVM rescue page with Claim/Refund flows.
  * RBTC added to refundable/rescue flows and faster background preimage processing for improved recovery.

* **Tests**
  * Large e2e test suite added and updated to cover RSK rescue/resume/refund scenarios and utilities.

* **Chores**
  * Updated RSK timeouts/config and added CI environment variable for e2e runs; submodule pointers updated.

* **Style**
  * Added a small text-spacing utility class.

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -440,7 +440,7 @@ const dict = {
         error_occurred: "An error occurred: {{ error }}",
 
         rsk_rescue_prompt:
-            "Refund where you were sending RBTC, resume where you were receiving RBTC of a swap that is not available in this browser’s swap history.",
+            "Use refund if you were sending RBTC, resume if you were receiving RBTC of a swap that is not available in this browser’s swap history.",
         rsk_rescue_refund_title: "Refund Swap",
         rsk_rescue_refund_explainer:
             "Connect your Rootstock wallet to scan for swaps that have expired and can be refunded.",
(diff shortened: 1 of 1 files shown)


src/utils/contractLogs.ts
```diff
@@ -165,4 +232,261 @@ const parseLockupEvent = (
...
+    worker?.terminate();
+    log.info(`Finished lockup event scanning`);
+}
+
+/**
+ * Finds the highest preimage key index used by an address.
+ * Scans blockchain events and derives preimages to find the highest index.
+ * Returns -1 if no events found or configuration is missing.
+ */
+export const getHighestKeyIndex = async (
```
michael1011: This function is unused

src/pages/RescueEvm.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+            if (useRif) {
+                transactionHash = await relayClaimTransaction(
+                    signer(),
+                    getEtherSwap(),
+                    currentPreimage.toString("hex"),
+                    Number(props.claimData.amount),
+                    props.claimData.refundAddress,
+                    Number(props.claimData.timelock),
+                );
```
michael1011: You can't just use RIF to claim. What you'd need to do is call `getSmartWalletAddress` and also scan for that address as claim address. Then you *have* to use RIF to claim. Else you *have* to actually use the signer

michael1011 (changes requested):
@kilrau check translations please

e2e/evm/rskRescue.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+
+const selectAndClickSwapItem = async (page: Page) => {
+    const swapListItem = page.locator(".swaplist-item").first();
+    await expect(swapListItem).toBeVisible();
+    await swapListItem.click();
+};
+
+test.describe("RSK Rescue", () => {
+    test.beforeEach(async ({ injectProvider }) => {
+        applyBoltzConfPatch();
```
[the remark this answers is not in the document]
kiocos: Correct! In this case, I should probably rename this function, because it only checks if the patch was already applied  and throws an error if not:
```
export const applyBoltzConfPatch = () => {
    try {
        execSync("git apply --check --reverse boltz.conf.patch", {
            stdio: "pipe",
        });
    } catch {
        throw new Error(`
            (!) This test requires boltz.conf.patch to be applied.
            It will fail without it due to the swap timeout being different from what's expected.

            Please, run "git apply boltz.conf.patch" to apply the patch (or manually update your boltz.conf with the patch's values),
            then restart your regtest containers.
        `);
    }
};
```

PR #1183, opened 2026-02-25
fix: exception when reloading during backup phase

kiocos:
Steps to reproduce on `main`:
1. On a new browser, create a swap;
2. Once you are asked to download a new Rescue Key, reload the page;
3. This will lead to infinite loading once the backup is done or an exception (if `assetReceive` is `RBTC`);

Since the necessary data to create the swap is lost on reload, we redirect the user to `/swap` once the backup is done.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented failed balance/gas lookups when wallet signer is unavailable for RBTC, falling back to safe defaults to avoid errors.

* **Improvements**
  * Added on-mount validation that redirects users back to swap setup when required info is missing, preventing incomplete swap flows.
  * Added support for Bolt12 Lightning offers and improved URL handling for Lightning amount/destination parsing.

* **Tests**
  * Added end-to-end coverage for backup flow, page reloads, and Bolt12 destination scenarios.

### DIFF
--- a/e2e/create.spec.ts
+++ b/e2e/create.spec.ts
@@ -271,7 +271,7 @@ test.describe("page reload during backup: redirect to /swap after completion", (
             ).toBeVisible();
 
             await completeBackup(page);
-            await expect(page).toHaveURL(/\/swap/);
+            await expect(page).toHaveURL(/\/swap$/);
         });
     });
 });


src/pages/Create.tsx
```diff
@@ -269,6 +275,22 @@ const Create = () => {
...
+        if (
+            creatingSwap() &&
+            ((onchainAddress() === "" && btcChains.includes(assetReceive())) ||
+                (signer() === undefined &&
+                    evmChains.includes(assetReceive())) ||
+                (lnurl() === "" &&
+                    bolt12Offer() === undefined &&
+                    assetReceive() === LN))
+        ) {
+            navigate("/");
```
michael1011: The comment says `/swap` and that's what we should do. Not redirect to `/`

PR #1188, opened 2026-02-27
refactor: simplify ref handling

kilrau:
Removed ref id from local storage altogether and always set to our defaults (read "hardcoded"). The previous approach of persisting in local storage without ever updating led to some old ref ids causing misleading data for some of our partners. 

## Summary by CodeRabbit

* **Refactor**
  * Referral handling now derives a runtime default; persisted mutable referral and its setter were removed. Creation flows and network calls automatically use the computed referral.
* **Chores**
  * Startup migration cleans up legacy referral storage and increments the storage version.
* **Tests**
  * End-to-end tests updated to follow the new referral initialization flow and root navigation.


src/utils/migration.ts
michael1011: Looking back at this, really bad idea to just run stuff every time the site is loaded. We can't just keep adding things to this. Start versioning

src/context/Global.tsx
```diff
@@ -170,14 +169,7 @@ const GlobalProvider = (props: { children: JSX.Element }) => {
...
 
-    const [ref, setRef] = makePersisted(
-        // eslint-disable-next-line solid/reactivity
-        createSignal(defaultReferral()),
-        {
-            name: referralIdKey,
-            ...stringSerializer,
-        },
-    );
+    const ref = () => defaultReferral();
```
michael1011: This should not be here anymore. It's effectively a const

src/components/CreateButton.tsx
```diff
@@ -404,7 +404,7 @@ const CreateButton = () => {
                             sendAmount(),
                             receiveAmount(),
                             invoice(),
-                            ref(),
+                            referral,
```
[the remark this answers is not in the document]
kilrau: removed, all call `getReferral()` internally now

PR #1328, opened 2026-04-10
fix: handle auto-claim of already claimed swap

kiocos:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cooperative-claim recovery: when a swap lockup output is already spent the app records the spend, updates stored claim transaction info, and avoids showing duplicate claim-failure UI.

* **New Features**
  * Added on-chain output spend detection to determine whether a swap output was spent and act accordingly.

* **Tests**
  * Added tests covering claim-failure recovery, persistence updates, and error-notification behavior.

### DIFF
--- a/src/context/Pay.tsx
+++ b/src/context/Pay.tsx
@@ -275,7 +275,7 @@ const PayProvider = (props: { children: JSX.Element }) => {
                     if (vout !== undefined) {
                         try {
                             log.debug(
-                                `checking for spent status of tx ${data.transaction.id}:${vout} from swap ${currentSwap.id}`,
+                                `swap ${currentSwap.id}: checking for spent status of tx output ${data.transaction.id}:${vout}`,
                             );
                             const outspend = await getTransactionOutSpend(
                                 currentSwap.assetReceive,
@@ -285,7 +285,7 @@ const PayProvider = (props: { children: JSX.Element }) => {
 
                             if (outspend?.spent && outspend.txid) {
                                 log.debug(
-                                    `lockup tx from swap ${currentSwap.id} was already claimed, updating claimTx id to ${outspend.txid}`,
+                                    `swap ${currentSwap.id}: lockup tx was already claimed, updating claimTx id to ${outspend.txid}`,
                                 );
                                 currentSwap.claimTx = outspend.txid;
                                 await setSwapStorage(currentSwap);
--- a/src/context/Pay.tsx
+++ b/src/context/Pay.tsx
@@ -12,11 +12,14 @@ import { hiddenInformation } from "src/components/settings/PrivacyMode";
 import { BTC, LBTC } from "src/consts/Assets";
 import { SwapType } from "src/consts/Enums";
 import { swapStatusPending, swapStatusSuccess } from "src/consts/SwapStatus";
+import { getTransactionOutSpend } from "src/utils/blockchain";
 import {
     claim,
     createSubmarineSignature,
     createTheirPartialChainSwapSignature,
+    findSwapOutputVout,
 } from "src/utils/claim";
+import { getTransaction } from "src/utils/compat";
 import { getPair } from "src/utils/helper";
 
 import {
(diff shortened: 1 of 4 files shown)


kilrau (changes requested):
Good approach! What's missing:
- Mirror the normal success path in the fallback by updating the active `swap()` signal too, not just storage. User shouldn't need to refresh the page "to fix it". Unless there is a good reason for not doing that?
- Let's wrap the new `getTransactionOutSpend()` call safely. If all the explorer lookups fail, this can now throw from inside the `catch` block and skip the normal `claim_fail` error handling.
- Add a test for regression

src/utils/claim.ts
```diff
@@ -320,6 +320,53 @@ const claimChainSwap = async (
     }
 };
 
+export const findSwapOutputVout = (
```
michael1011, line 370: That's duplicate code

src/context/Pay.tsx
```diff
@@ -254,6 +257,41 @@ const PayProvider = (props: { children: JSX.Element }) => {
                     }),
                 );
             } catch (e) {
+                if (
+                    typeof e === "string" &&
+                    e.includes("bad-txns-inputs-missingorspent")
```
michael1011: Is it the same error for Esplora and Mempool on both, the mainchain and Liquid?

src/context/Pay.tsx
```diff
@@ -254,6 +257,41 @@ const PayProvider = (props: { children: JSX.Element }) => {
...
+                    if (vout !== undefined) {
+                        const outspend = await getTransactionOutSpend(
+                            currentSwap.assetReceive,
+                            data.transaction.id,
+                            vout,
+                        );
+
+                        if (outspend?.spent && outspend.txid) {
+                            log.debug(
+                                `tx ${data.transaction.id}:${vout} was already claimed, updating claimTx id to ${outspend.txid}`,
```
michael1011: Include the swap id in the logs

src/context/Pay.tsx
```diff
@@ -254,6 +257,41 @@ const PayProvider = (props: { children: JSX.Element }) => {
...
+                        currentSwap.assetReceive,
+                    ).fromHex(data.transaction.hex);
+                    const vout = findSwapOutputVout(
+                        deriveKey,
+                        currentSwap as ReverseSwap | ChainSwap,
+                        lockupTx,
+                    );
+
+                    if (vout !== undefined) {
+                        const outspend = await getTransactionOutSpend(
```
michael1011: Is not error handled

src/context/Pay.tsx
```diff
@@ -254,6 +257,41 @@ const PayProvider = (props: { children: JSX.Element }) => {
...
+                            data.transaction.id,
+                            vout,
+                        );
+
+                        if (outspend?.spent && outspend.txid) {
+                            log.debug(
+                                `tx ${data.transaction.id}:${vout} was already claimed, updating claimTx id to ${outspend.txid}`,
+                            );
+                            currentSwap.claimTx = outspend.txid;
+                            await setSwapStorage(currentSwap);
```
michael1011: The swap in the pay context is not updated

michael1011 (changes requested):
And no test

src/context/Pay.tsx
kiocos: ~yes~

edit: actually... not really, pushed other strings from Core:
```
bilingual_str TransactionErrorString(const TransactionError err)
{
    switch (err) {
        case TransactionError::OK:
            return Untranslated("No error");
        case TransactionError::MISSING_INPUTS:
            return Untranslated("Inputs missing or spent");
        case TransactionError::ALREADY_IN_UTXO_SET:
            return Untranslated("Transaction outputs already in utxo set");
        case TransactionError::MEMPOOL_REJECTED:
            return Untranslated("Transaction rejected by mempool");
        case TransactionError::MEMPOOL_ERROR:
            return Untranslated("Mempool internal error");
        case TransactionError::MAX_FEE_EXCEEDED:
            return Untranslated("Fee exceeds maximum configured by user (e.g. -maxtxfee, maxfeerate)");
        case TransactionError::MAX_BURN_EXCEEDED:
            return Untranslated("Unspendable output exceeds maximum configured by user (maxburnamount)");
        case TransactionError::INVALID_PACKAGE:
            return Untranslated("Transaction rejected due to invalid package");
    } // no default case, so the compiler can warn about missing cases
    assert(false);
}
```

src/utils/claim.ts
```diff
@@ -36,12 +36,79 @@ import type { ChainSwap, ReverseSwap, SubmarineSwap } from "./swapCreator";
 import { getRelevantAssetForSwap } from "./swapCreator";
 import { createMusig, hashForWitnessV1, tweakMusig } from "./taproot/musig";
 
+type ClaimSwap = ReverseSwap | ChainSwap;
+type ClaimPrivateKey = ReturnType<typeof parsePrivateKey>;
+type ClaimSwapTree = ReturnType<typeof SwapTreeSerializer.deserializeSwapTree>;
+type ClaimKeyAgg = ReturnType<typeof createMusig>;
+type ClaimTweaked = ReturnType<typeof tweakMusig>;
```
michael1011: Isn't there strong types for all these things already?

src/utils/claim.ts
```diff
@@ -36,12 +36,79 @@ import type { ChainSwap, ReverseSwap, SubmarineSwap } from "./swapCreator";
 import { getRelevantAssetForSwap } from "./swapCreator";
 import { createMusig, hashForWitnessV1, tweakMusig } from "./taproot/musig";
 
+type ClaimSwap = ReverseSwap | ChainSwap;
```
michael1011:
```suggestion
type ClaimableSwap = ReverseSwap | ChainSwap;
```

kilrau (approved):
tACK (only merge after addressing Michael's review)

main: removing UTXO claim tx leads to `Failed to claim swap: sWaPiD` error on every page load

this branch: redisovers UTXO claim tx and moves swap back to success state with claim tx button. works on all pages as expected, no need to navigate to swap page.

https://github.com/user-attachments/assets/4a46564f-2b4a-462c-bc98-8164e38d67e2

michael1011 (approved):

PR #1338, opened 2026-04-15
fix: make EVM address comparison case insensitive

kiocos:
## Summary by CodeRabbit

* **Bug Fixes**
  * Address matching in contract transactions is now case-insensitive, ensuring the "Send" action appears correctly regardless of address casing.

* **Refactor**
  * Wallet connection setup has been consolidated to improve reliability and reduce duplicated logic during wallet connect and chain changes.

* **Tests**
  * Added tests validating wallet/address matching and the visibility of the "Send" action.


michael1011 (changes requested):
Has that ever been a problem?

michael1011 (approved):

PR #1343, opened 2026-04-15
fix: block 0 amount receive

### LINKED ISSUE #1333: Strictly block swap creation if receive amount 0 or quote still loading
Currently one can create USDT swaps with a `receiveAmount` of `0` and/or quote still loading:

Image

Leading to a very annoying state:

Image

Which we currently only allow to refund after swap timeout.

jackstar12:
closes: https://github.com/BoltzExchange/boltz-web-app/issues/1333

## Release Notes

* **Bug Fixes**
  * Added validation to prevent swap creation when a send amount is entered but the quoted receive amount is zero, with a clear error message prompting users to increase the send amount.

* **Localization**
  * Added translated error message across all supported languages.

### DIFF
--- a/src/i18n/i18n.ts
+++ b/src/i18n/i18n.ts
@@ -706,8 +706,7 @@ const dict = {
         retry: "Erneut versuchen",
         error_no_quote:
             "Ein Angebot konnte nicht abgerufen werden. Bitte prüfe deine Verbindung und versuche es erneut.",
-        error_zero_quote:
-            "Der angezeigte Empfangsbetrag ist null. Bitte erhöhe den Sendebetrag und versuche es erneut.",
+        error_zero_quote: "Sendebetrag erhöhen",
         waiting_for_oft: "Warte auf OFT",
         dex_quote_changed: "DEX-Kurs hat sich geändert",
         zero_conf: "Zero-Conf",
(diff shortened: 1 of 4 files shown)


kiocos (commented):
This error message is fine to me, gotta update the english string to match it 

Also remove the end "." in all languages

Image

src/i18n/i18n.ts
```diff
@@ -705,6 +706,8 @@ const dict = {
         retry: "Erneut versuchen",
         error_no_quote:
             "Ein Angebot konnte nicht abgerufen werden. Bitte prüfe deine Verbindung und versuche es erneut.",
+        error_zero_quote:
+            "Der angezeigte Empfangsbetrag ist null. Bitte erhöhe den Sendebetrag und versuche es erneut.",
```
michael1011:
```suggestion
            "Der angezeigte Empfangsbetrag ist Null. Bitte erhöhe den Sendebetrag und versuche es erneut.",
```

src/i18n/i18n.ts
```diff
@@ -211,6 +211,7 @@ const dict = {
         retry: "Retry",
         error_no_quote:
             "A quote could not be obtained. Please check your connection and try again.",
+        error_zero_quote: "Increase send amount",
```
michael1011, line 214: As Caio said, that's inconsistent

michael1011 (changes requested):

src/i18n/i18n.ts
jackstar12: Translation fuckup

michael1011 (approved):

PR #1346, opened 2026-04-16
chore: optimize solana rpc calls

jackstar12:
## Summary by CodeRabbit

* **Refactor**
  * Unified scattered Solana caching into a single prefix-based cache with one clear operation; replaced ad-hoc cache handling across the codebase.
  * Routed repeated on-chain reads through the shared cache to simplify logic and reduce duplicated retrieval flows.

* **New Features**
  * Added a general-purpose caching utility with conditional retention and in-flight deduplication.
  * Added cached account-info retrieval to cut repeated RPC calls.

* **Tests**
  * Updated and added tests covering cache semantics, dedupe, conditional retention, and invalidation.

### DIFF
--- a/src/utils/chains/solana.ts
+++ b/src/utils/chains/solana.ts
@@ -8,17 +8,83 @@ import { config } from "../../config";
 import { NetworkTransport } from "../../configs/base";
 import lazySolana from "../../lazy/solana";
 import { formatError } from "../errors";
-import { constructRequestOptions } from "../helper";
 import { requireRpcUrls } from "../provider";
 
 export const solanaAddressLength = 32;
 export const solanaTokenAccountSize = 165;
 export const solanaAtaRentExemptLamports = 2_039_280n;
-const solanaGetAccountInfoMethod = "getAccountInfo";
-const solanaRpcProbeTimeout = 5_000;
-const solanaTokenAccountExistsCache = new Set<string>();
-const solanaConnectionCache = new Map<string, Promise<Connection>>();
-const solanaRentExemptBalanceCache = new Map<string, Promise<bigint>>();
+const rpcProbeTimeout = 5_000;
+
+type AccountInfo = Awaited<ReturnType<Connection["getAccountInfo"]>>;
+
+type CacheOptions<T> = {
+    shouldCache?: (value: T) => boolean;
+};
+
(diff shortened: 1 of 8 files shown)


src/utils/chains/solana.ts
```diff
@@ -8,17 +8,83 @@ import { config } from "../../config";
...
+            const cacheKey = getCacheKey(prefix, key);
+            const cached = entries.get(cacheKey);
+            if (cached !== undefined) {
+                return cached as Promise<T>;
+            }
+
+            return set(cacheKey, create(), options);
+        },
+    };
+};
```
michael1011: Why can't we create a new utility class and have to hack around with hoisted variables?

src/utils/chains/solana.ts
```diff
@@ -8,17 +8,83 @@ import { config } from "../../config";
...
+    const set = <T>(
+        key: string,
+        value: Promise<T>,
+        { shouldCache }: CacheOptions<T> = {},
+    ): Promise<T> => {
+        const cached = value
+            .then((resolved) => {
+                if (shouldCache !== undefined && !shouldCache(resolved)) {
+                    entries.delete(key);
+                }
```
michael1011: When you should not cache you delete cached values?

src/utils/chains/solana.ts
```diff
@@ -44,16 +110,8 @@ export const encodeSolanaRecipient = (recipient: string): string =>
...
-};
-
-export const clearSolanaConnectionCache = () => {
-    solanaConnectionCache.clear();
-};
-
-export const clearSolanaRentExemptBalanceCache = () => {
-    solanaRentExemptBalanceCache.clear();
+export const clearSolanaCache = () => {
+    cache.clear();
```
michael1011: I ***hate*** that we export these dirty test helpers everywhere

michael1011 (changes requested):

src/utils/chains/solana.ts
jackstar12: We put the promise immediately in the cache even before it resolves. If it resolves to a value we dont wanna keep we rm - e.g. for the solana token account creation requirement

michael1011 (approved):

PR #1355, opened 2026-04-17
refactor: generic bridge interface

michael1011:
## Summary by CodeRabbit

* **New Features**
  * Introduces a generalized "bridge" system replacing legacy OFT-specific flows.

* **Improvements**
  * Asset/network selection now supports canonical bridge variants and explicit network selection for bridged assets.
  * UI and flows (fees display, messaging/transfer fee handling, send/claim/refund flows, waiting screens, block explorer links) are bridge-aware.
  * RPC/provider handling improved with per-network overrides and immutable RPC lists.

* **Data Migration**
  * Storage version bumped to v7 with automatic migration of legacy swaps to the new bridge shape.

* **Tests**
  * Updated and added tests covering bridge registry, drivers, routing, and bridge flows.

* **Chores**
  * Added new environment variables for RPCs and bridge-related services.


src/components/BlockExplorerLink.tsx
```diff
@@ -32,11 +37,9 @@ const ChainSwapLink = (props: {
...
                     asset={asset()}
                     txId={props.swap().claimTx}
-                    explorer={
-                        props.swap().oft !== undefined && isEvmAsset(asset())
-                            ? ExplorerKind.LayerZero
-                            : undefined
-                    }
+                    explorer={bridgeRegistry.getExplorerKind(
+                        getPostBridgeDetail(props.swap().bridge),
+                    )}
```
kiocos, line 42: if `driver` is undefined, it will throw
```
    public requireDriverForAsset = (asset: string): BridgeDriver => {
        const driver = this.getDriverForAsset(asset);
        if (driver === undefined) {
            throw new Error(`No bridge driver found for asset ${asset}`);
        }

        return driver;
    };
```
not sure how the UI would behave in such case, ig we should make this fail more "gracefully"

src/utils/bridge/types.ts
```diff
@@ -0,0 +1,90 @@
...
+    OftTransaction,
+    OftTransportClient,
+    OftTransportRunner,
+    SendParam,
+} from "../oft/types";
+import type { Provider } from "../provider";
+
+export type BridgeRoute = OftRoute;
+
+export type BridgeDetail = BridgeRoute & {
```
kiocos, line 24: this is declared twice

src/components/BlockExplorerLink.tsx
michael1011: In that case it's a grave misconfiguration that should throw and blow up

PR #1367, opened 2026-04-22
fix: TrustWallet on iOs v2

kiocos:
Fixes:
- the `eth_signTypedData_v4` payload formatting which was being rejected by TrustWallet:
```
error Transaction failed {
  "__error": true,
  "name": "Error",
  "message": "could not coalesce error (
    error={
      \"code\": \"UNKNOWN_ERROR\",
      \"message\": \"could not coalesce error (
        error={
          \\\"code\\\": -32000,
          \\\"message\\\": \\\"The string did not match the expected pattern.\\\"
        },
        payload={
          \\\"id\\\": 2,
          \\\"jsonrpc\\\": \\\"2.0\\\",
          \\\"method\\\": \\\"eth_signTypedData_v4\\\",
          ...
        }
      )\"
    }
  )"
}
```
- sends Trust Wallet signing requests through the WalletConnect provider, fixing:
```
could not coalesce error
(error={ "code": -32601, "message": "Method not found" }, code=UNKNOWN_ERROR)
```

## Summary by CodeRabbit

* **New Features**
  * Better typed-data signing for Trust Wallet on iOS via WalletConnect.

* **Improvements**
  * More reliable EVM session handling with persisted chain/address state and safer clearing when provider or chain info is unavailable.
  * Dynamic network detection for EVM, Solana, and Tron instead of hardcoded network names.
  * Routing of certain read-only EVM RPC calls to cached JSON‑RPC senders for improved stability.

### DIFF
--- a/src/context/Web3.tsx
+++ b/src/context/Web3.tsx
@@ -12,7 +12,11 @@ import {
     Contract,
     type InterfaceAbi,
     JsonRpcSigner,
+    type TypedDataDomain,
+    TypedDataEncoder,
+    type TypedDataField,
     Wallet,
+    getAddress,
 } from "ethers";
 import log from "loglevel";
 import type { Accessor, JSXElement, Setter } from "solid-js";
@@ -47,6 +51,7 @@ import { getContracts } from "../utils/boltzClient";
 import type { HardwareSigner } from "../utils/hardware/HardwareSigner";
 import LedgerSigner from "../utils/hardware/LedgerSigner";
 import TrezorSigner from "../utils/hardware/TrezorSigner";
+import { isIos } from "../utils/helper";
 import {
     createAssetProvider,
     getRpcUrls,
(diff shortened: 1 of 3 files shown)


src/context/Web3.tsx
```diff
@@ -569,6 +611,20 @@ const Web3SignerProvider = (props: {
             return;
         }
 
+        if (
+            activeSigner?.rdns === walletConnectRdns &&
+            isIos() &&
```
michael1011: This does not just affect iOS

michael1011 (changes requested):
Switch Network button does not work with Trust Wallet on Android

src/context/Web3.tsx
kiocos: Also works on Android now.

I tested TrustWallet on desktop too: works with the extension, but not with WalletConnect + TW (WalletConnect doesn't find recognize TW is installed). I think this is unrelated and their integration issue, so leaving it untouched for now

michael1011 (approved):
Not the cleanest thing I have seen in my life but works fine on Trust Wallet on Android

PR #1378, opened 2026-04-24
fix: allow entering amounts for invalid pairs

### LINKED ISSUE #1374: `Invalid Pair` doesn't allow entering amounts
In this state, I can only enter one digit. We should allow entering amounts in all invalid pair states in all min/max 0 states the same way we do for valid pair states.

Image

kiocos:
Closes https://github.com/BoltzExchange/boltz-web-app/issues/1374


kilrau (approved):
tACK

michael1011 (approved):

PR #1382, opened 2026-04-24
CCTP

### LINKED ISSUE #1249: USDC: CCTP
https://www.circle.com/cross-chain-transfer-protocol

### LINKED ISSUE #1383: Mint USDC in lockup transaction
When sending USDC from a chain that is not Arbitrum, we don't need the Circle Forwarding Service and can just do the mint ourselves in the lockup transaction as soon as the attestation is available

Followup for https://github.com/BoltzExchange/boltz-web-app/pull/1382

michael1011:
Closes https://github.com/BoltzExchange/boltz-web-app/issues/1249
Closes #1383

## Release Notes

* **New Features**
  * Added USDC asset support with CCTP bridge integration
  * Expanded network support: Avalanche, Base, Codex, Linea, Plume, Sonic, World Chain, and XDC
  * Implemented Circle CCTP v2 protocol for USDC transfers

* **Bug Fixes**
  * Improved messaging fee handling in bridge operations

* **Chores**
  * Updated boltz-core dependency to v4.0.4
  * Enhanced configuration structure for bridge types


src/utils/bridge/CctpBridgeDriver.ts
```diff
@@ -0,0 +1,569 @@
...
+    ): Promise<BridgeReceivedEvent | undefined> => {
+        void contract;
+        void contractAddress;
+
+        const decoded = decodeCctpGuid(guid);
+        if (decoded === undefined) {
+            throw new Error(`invalid CCTP guid: ${guid}`);
+        }
+
+        const forwardTxHash = await getCctpForwardTxHash(
```
kiocos, line 306: if this throws, it will break the retry loop

i guess we should keep retrying on transient errors, like network errors, 5xx and alike

src/utils/cctp/fee.ts
```diff
@@ -0,0 +1,154 @@
...
+            ),
+        };
+    } finally {
+        clearTimeout(requestTimeout);
+    }
+};
+
+export const cctpFeeBpsDenominator = 10_000n * cctpFeeScale;
+
+export const getCctpFee = async (
```
jackstar12: We shouldn't be caching this fee. As per circle's docs, this should be fetched immediately before initiating the transfer:

https://developers.circle.com/cctp/howtos/transfer-usdc-with-forwarding-service#step-1-get-cctp-fees-from-the-api
 
> Circle recommends selecting the med fee level or higher from the forwardFee object in the API response. Note that forwardFee values fluctuate based on destination chain gas prices. Make the query immediately before initiating your transfer.

src/utils/cctp/fee.ts
jackstar12: This is a pure wrapper around `fetchCctpFee` now, can be removed

src/utils/oft/registry.ts
```diff
@@ -72,7 +72,7 @@ export const getOftChain = async (
 
     const assetConfig = config.assets?.[asset];
     const meshKind =
-        getAssetBridge(route.sourceAsset)?.kind !== undefined
+        getAssetBridge(route.sourceAsset)?.kind === BridgeKind.Oft
```
jackstar12: shouldn't we treat a non-oft bridge asset entering this code path as an error?  we could embed that check in `getBridgeMesh` and throw from there, covering all the callers that do exactly this check

kilrau (approved):
tACK

PR #1388, opened 2026-04-25
fix: Tron mobile wallet detection

### LINKED ISSUE #1386: Use mnemonic backup for Tron sends on mobile
Just like [EVM sends on mobile](https://github.com/BoltzExchange/boltz-web-app/blob/b2600d6de7800a667bffc6cd20f654593f982b9b/src/components/BackupDownloadContent.tsx#L19), Tron sends on mobile would need to use mnemonic-based backup as Tron Mobile wallets like TronLink do not allow downloading backup files.

Follow-up of https://github.com/BoltzExchange/boltz-web-app/pull/1354.

michael1011:
Closes https://github.com/BoltzExchange/boltz-web-app/issues/1386

## Summary by CodeRabbit

* **Improvements**
  * Enhanced detection of available wallet types (EVM and TRON) on mobile browsers
  * Updated backup recovery and wallet connection flows to accurately reflect which wallets are accessible to users

### DIFF
--- a/src/context/Web3.tsx
+++ b/src/context/Web3.tsx
@@ -71,6 +71,9 @@ declare global {
     interface Window {
         // @ts-expect-error - conflicts with @reown/appkit-utils non-optional declaration
         ethereum?: EIP1193Provider;
+        tron?: unknown;
+        tronLink?: unknown;
+        tronWeb?: unknown;
     }
 }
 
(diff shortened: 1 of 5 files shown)


michael1011:
@coderabbitai review

src/context/Web3.tsx
```diff
@@ -71,6 +71,9 @@ declare global {
     interface Window {
         // @ts-expect-error - conflicts with @reown/appkit-utils non-optional declaration
         ethereum?: EIP1193Provider;
+        tron?: unknown;
+        tronLink?: unknown;
+        tronWeb?: unknown;
```
jackstar12, line 76: So the tron devs couldn't agree on a name here?

michael1011: These are all slightly different versions of the same thing. All we care about is checking if its a custom mobile browser for which any of those not being undefined is good enough of a heuristic

jackstar12 (approved):

PR #1392, opened 2026-04-27
fix: delete button css

kiocos:
fixes a regression which made the delete button yellow
image


michael1011 (approved):
Regression from https://github.com/BoltzExchange/boltz-web-app/commit/eeb0e3f

PR #1393, opened 2026-04-27
fix: wait for tx confirmation for oft eta

### LINKED ISSUE #1385: OFT Timer should wait for tx confirmation
If blockchain confirmation takes some time, our OFT ETA is off compared to LZs. Can be easily reproduced, e.g. sending `USDT0-POL` into a swap.
Image
Image

kiocos:
## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable confirmation waits for OFT transfers: respects cancellations, avoids stale or incorrect ETA updates, and prevents state changes after leaving the screen.
* **Tests**
  * Added unit tests for the confirmation-wait flow and ETA display to guard against regressions.

### DIFF
--- a/src/components/WaitForBridge.tsx
+++ b/src/components/WaitForBridge.tsx
@@ -5,7 +5,7 @@ import { BridgeKind } from "../configs/base";
 import { useGlobalContext } from "../context/Global";
 import { bridgeRegistry } from "../utils/bridge";
 import { formatError } from "../utils/errors";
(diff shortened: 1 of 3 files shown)


kilrau (approved):
image


There still is a discrepancy to layerzero.com's web page as you mentioned but funnily our ETA is better than theirs now. Our timer was exactly right - hit 1s when we started dispatching the Lightning invoice. Invoice payment completed before the timer on layerzero.com showed the OFT transfer as completed.

michael1011 (approved):

PR #1398, opened 2026-04-29
fix: Blockstream logo

kiocos:
## Summary by CodeRabbit

* **Style**
  * Refined typography and spacing behavior in the application layout to enhance visual consistency and overall presentation.

### DIFF
--- a/src/style/hero.scss
+++ b/src/style/hero.scss
@@ -259,22 +259,19 @@ h3 {
 }
 .blockstreamApp {
     position: relative;
-    padding-left: 64px;
+    padding-left: 62px;
     text-decoration: none;
     font-size: 18px;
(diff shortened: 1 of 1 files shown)


kilrau (changes requested):
Can you show me where this is broken? I just checked prod and beta deployment on different desktop and mobile browsers and looks fine to me.

Also, if it is broken somewhere - let's fix it instead of removing "App" because it is called "Blockstream App".

Image

kiocos:
Before (Firefox):
image

Now:
image

I also tested on a Chrome-based browser, looks fine on it too.

kilrau (approved):
Confirmed: broken on Firefox only
image

And confirmed this PR fixes it:
image

PR #1405, opened 2026-04-29
feat: make OFT ETA configurable via env

kiocos:
Requires `is_production_deploy = os.environ.get("VERCEL_ENV") == "production"` to be applied in production

## Summary by CodeRabbit

* **New Features**
  * Per-network OFT ETA can now be configured via environment variables for 25 networks.

* **Chores**
  * Startup now validates OFT ETA values in production and fails on missing/invalid numeric entries.
  * Runtime reads OFT ETA from environment, allowing dynamic overrides and logging missing/invalid values.

* **Documentation**
  * .env.sample updated with clearer usage comments; .env/local override guidance added and .env.local is ignored by VCS.

PR #1406, opened 2026-04-29
feat: make sendable usdt0 variants configurable via env

kiocos:
## Summary by CodeRabbit

* **New Features**
  * Added per-chain environment flags to enable or disable sending of USDT0 on individual networks.
* **Chores**
  * Updated sample environment configuration to include the new per-chain overrides and ensure defaults remain unchanged.


michael1011 (approved):

src/configs/usdt0.ts
```diff
@@ -54,6 +54,25 @@ import { AssetKind } from "src/consts/AssetKind";
 import { solanaMinGasTopUpLamports } from "src/consts/Solana";
 import { getExplorerId } from "src/utils/explorer";
 
+const envCanSend = (
+    envValue: string | undefined,
```
michael1011, line 58: Can we make this case insensitive and trim?

michael1011 (approved):

src/configs/usdt0.ts
```diff
@@ -101,7 +124,7 @@ const createUsdt0VariantAsset = (variant: Usdt0Variant): Asset => {
 export const usdt0Variants = [
     {
         asset: "USDT0-ETH",
-        canSend: true,
+        canSend: envCanSend(import.meta.env.VITE_USDT0_ETH_CAN_SEND, true),
```
jackstar12: Couldn't we do this one dynamically? Smth like `envCanSend(import.meta.env["VITE_USDT0_{asset}_CAN_SEND")` in the `createUsdt0VariantAsset` helper

jackstar12: Like here: https://github.com/BoltzExchange/boltz-web-app/pull/1405/changes#diff-d44ce25b9100b39f0246129fbf16235b8c392ea39bb89972ac3fe143c1750b91R5

jackstar12 (dismissed):

src/configs/usdt0.ts
kiocos: Just validated we can't: 
>During production, these env variables are statically replaced. It is therefore necessary to always reference them using the full static string. For example, dynamic key access like import.meta.env[key] will not work.

https://v2.vitejs.dev/guide/env-and-mode.html#production-replacement

I'll double check the example you linked

kiocos: @jackstar12 opted for declaring a map and re-using it in both places:
https://github.com/BoltzExchange/boltz-web-app/pull/1406/changes/0ccdf40a3a556cf1b4771f09dc8559847255fdbc
https://github.com/BoltzExchange/boltz-web-app/pull/1405/changes/0b389253f07a5402b674212956d9af63c33d3843

michael1011 (approved):

PR #1427, opened 2026-05-04
feat: handle abandoned sol txes

jackstar12:
Close #1316

## Summary by CodeRabbit

* **Improvements**
  * Bridge transactions now include optional confirmation metadata (e.g., Solana blockhash) for more reliable tracking across networks.
  * Confirmation and polling logic enhanced for Solana, Tron and EVM flows to better detect expired/dropped sends and react earlier.

* **Bug Fixes**
  * Abandoned/failed bridge sends now remove stale transaction data and persist relevant details consistently.

* **Tests**
  * Added and updated tests covering Solana dropped-send scenarios and metadata handling.

### DIFF
--- a/src/components/SendToBridge.tsx
+++ b/src/components/SendToBridge.tsx
@@ -613,13 +613,17 @@ const SendToBridge = (props: {
             return;
         }
         if (currentSwap.bridge !== undefined) {
-            const bridge = { ...currentSwap.bridge };
-            delete bridge.details;
-            currentSwap.bridge = {
-                ...bridge,
+            const bridge = {
+                ...currentSwap.bridge,
                 txHash: tx.hash,
-                ...(tx.details === undefined ? {} : { details: tx.details }),
             };
+            if (tx.details === undefined) {
+                delete bridge.details;
+            } else {
+                bridge.details = tx.details;
+            }
+
+            currentSwap.bridge = bridge;
         }
 
         setSwap(currentSwap);
--- a/src/components/SendToBridge.tsx
+++ b/src/components/SendToBridge.tsx
@@ -20,7 +20,7 @@ import {
 } from "../context/Web3";
 import type { DictKey } from "../i18n/i18n";
 import WalletConnectProvider from "../utils/WalletConnectProvider";
-import { bridgeRegistry } from "../utils/bridge";
+import { type BridgeTransaction, bridgeRegistry } from "../utils/bridge";
 import { getTronTokenAllowance } from "../utils/oft/oft";
 import type { OftTransportClient } from "../utils/oft/types";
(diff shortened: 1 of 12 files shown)


src/utils/bridge/details.ts
```diff
@@ -0,0 +1,7 @@
+export type SolanaDetails = {
+    blockhash: string;
+};
+
+export type BridgeDetails = Record<string, unknown> & {
```
michael1011:
```suggestion
export type BridgeDetails = {
```
Let's not throw in Records with unknowns

src/components/SwapExecutionWorker.tsx
```diff
@@ -524,10 +524,47 @@ export const SwapExecutionWorker = () => {
...
+        if (solana === undefined) {
+            return false;
+        }
+
+        const blockhashStatus = await connection.isBlockhashValid(
+            solana.blockhash,
+            {
+                commitment: "confirmed",
+            },
+        );
```
michael1011, line 572: This can still race. How about we check if the block hash is valid and if it's not *and* the transaction was not included, then we can abandon

src/components/SendToBridge.tsx
```diff
@@ -598,12 +592,15 @@ const SendToBridge = (props: {
...
         const currentSwap = await getSwap(props.swapId);
         if (currentSwap?.bridge !== undefined) {
+            const bridge = { ...currentSwap.bridge };
+            delete bridge.details;
             currentSwap.bridge = {
-                ...currentSwap.bridge,
-                txHash,
+                ...bridge,
+                txHash: tx.hash,
+                ...(tx.details === undefined ? {} : { details: tx.details }),
```
michael1011: This reads horribly

michael1011 (changes requested):

src/components/SwapExecutionWorker.tsx
jackstar12: How can it race? It can happen that the initial get signature call returns null but in that case the block hash will return valid any time. If the signature doesn't show up until the block hash isn't valid anymore we can abandon.

michael1011: 1. Get the tx status
2. Process the response
3. Get if the block range is still valid
4. Process the response

If the block range becomes invalid after getting the tx status, you abandon the tx but it could have been included in the window of time between 1 and 3

jackstar12: > If the block range becomes invalid after getting the tx status

Thats a big if. If it wasn't included by the time the blockhash has become invalid (which practically takes ~60-90 seconds according to solana docs) it's pretty certain that it simply isn't valid. Guarding against that doesn't hurt though I guess.

michael1011 (approved):

PR #1444, opened 2026-05-09
refactor: pre bridge logs chunking

### LINKED ISSUE #1441: Scan Arbitrum logs in chunks
In `getEvmOftReceivedEventByGuid`, we currently scan with:
```
    const logs = await provider.getLogs({
        address: getAddress(contractAddress),
        event: getAbiItem({
            abi: oftAbi,
            name: "OFTReceived",
        }),
        args: { guid },
        fromBlock: 0n,
        toBlock: "latest",
    });
```
we shouldn't do a single GET from `0x0` to `latest`. Rather, we should scan it in chunks, like we do in `contractLogs.ts`, for example.

michael1011:
Closes #1441

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bridge receipt polling resilience by implementing block snapshot tracking and reorg-tolerance windows for better handling of blockchain reorganizations.
  * Optimized event lookup efficiency by introducing batched block-range scanning strategy, replacing previous full-chain query approach.

* **Tests**
  * Added comprehensive test coverage for block-range batching logic, event scanning scenarios, and edge case handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/BoltzExchange/boltz-web-app/pull/1444)

### DIFF
--- a/src/utils/oft/evm.ts
+++ b/src/utils/oft/evm.ts
@@ -13,6 +13,7 @@ import {
 
 import { NetworkTransport } from "../../configs/base";
 import type { Signer } from "../../context/Web3";
+import { generateBlockRangeBatches } from "../evmBlockRanges";
 import type {
     MsgFee,
     OftFeeDetail,
@@ -23,6 +24,15 @@ import type {
     OftTransportClient,
     SendParam,
 } from "./types";
+
+// Cap the initial scan window for OFT receive lookups. The function is
+// polled while waiting for a bridge receive after the source send, so the
+// matching event is always recent. ~24h on Arbitrum at ~0.25s/block.
+const maxLookbackBlocks = 350_000;
+
+// Per-chunk window for parallel log scanning. ~12h on Arbitrum at ~0.25s/
+// block — two chunks cover the full lookback in a single parallel batch.
+const scanIntervalBlocks = 175_000;
 
 export type EvmOftTransportClient = OftTransportClient & {
     abi: typeof oftAbi;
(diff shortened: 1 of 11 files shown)


src/utils/oft/evm.ts
```diff
@@ -216,51 +226,75 @@ export const getEvmOftReceivedEvent = (
...
-        eventName: "OFTReceived",
-        logs,
-    })[0];
+    const address = getAddress(contractAddress);
+    const event = getAbiItem({ abi: oftAbi, name: "OFTReceived" });
+    const latestBlock = Number(await provider.getBlockNumber());
+    const minBlock =
+        options?.fromBlock !== undefined
+            ? Math.max(0, Math.min(Number(options.fromBlock), latestBlock))
+            : Math.max(0, latestBlock - maxLookbackBlocks);
```
kiocos, line 244: This would prevent the user from rescuing funds on a block older than `latestBlock - maxLookbackBlocks`.

I think we should keep scanning in chunks, but preserve the behavior of scanning all the way back to 0.

```suggestion
            : 0;
```

kiocos (changes requested):

src/utils/oft/evm.ts
michael1011: This is not the rescue page. If your swap expired, this function should not care anymore. Which is why we have a lookback period. RPC requests cost money, and we should not waste them on impossible situations

kiocos (approved):

PR #1451, opened 2026-05-10
Swaps package

michael1011:


packages/boltz-swaps/package.json
```diff
@@ -0,0 +1,194 @@
...
+        "prettier:write": "bun run prettier -- --write",
+        "prettier:check": "bun run prettier -- --check",
+        "lint": "bunx eslint --max-warnings 0 --cache src",
+        "check": "bun run prettier:write && bun run lint -- --fix && bun run tsc",
+        "generate:evm-abis": "bun run scripts/generate-evm-abis.ts",
+        "generate:solana-oft": "codama run solana-oft",
+        "generate:solana-cctp": "codama run solana-cctp --idl src/generated/solana-cctp-token-messenger-minter-v2-idl.json",
+        "generate:fix-codama-imports": "bun run scripts/fix-codama-imports.ts",
+        "generate": "bun run generate:evm-abis && bun run generate:solana-oft && bun run generate:solana-cctp && bun run generate:fix-codama-imports",
+        "postinstall": "bun run generate",
```
kiocos: Shouldn't we generate this at our main `package.json`, so we can have the generate files published and shipped under `dist` as well? I think would be the expected default for npm packages

kiocos (approved):

PR #1465, opened 2026-05-15
Extend package

michael1011:


packages/boltz-swaps/src/route.ts
```diff
@@ -0,0 +1,469 @@
...
+    route: BridgeRoute;
+    driver: BridgeDriver;
+};
+
+type LegPlan = {
+    chainSwap: {
+        from: string;
+        to: string;
+        pair: ChainPairTypeTaproot;
+    };
```
jackstar12: How do we wanna add in support for other swap types here in the future? I wouldn't like the idea of three different fields for each swap type.

packages/boltz-swaps/src/sdk.ts
```diff
@@ -0,0 +1,406 @@
...
+    };
+    readonly swap: {
+        chain: {
+            create(
+                args: ChainSwapCreateArgs<A>,
+            ): Promise<ChainSwapCreatedResponse>;
+            transactions: typeof getChainSwapTransactions;
+            newQuote: typeof getChainSwapNewQuote;
+            acceptNewQuote: typeof acceptChainSwapNewQuote;
+        };
```
jackstar12, line 184: Here again - let's make the creation a swap-agnostic interface. Unless we wanna keep some low-level primitives, but let's make the user-facing interface mostly swap-type agnostic. I don't wanna care about the fact that when sending from LN I have to create a different swap type.

src/utils/evmTransaction.ts
```diff
@@ -191,110 +116,23 @@ export const claimAsset = async (
     timeoutBlockHeight: number,
     destination: Address,
     signer: Accessor<Signer>,
-    getGasAbstractionSigner: Signer,
+    gasAbstractionSigner: Signer,
     etherSwap: EtherSwapContract,
     erc20Swap: Erc20SwapContract,
```
jackstar12: These params feel like gettin out of hand. Let's wrap some of these in a seperate type, maybe there's some potential for re-use here (at the boltz-swaps level)

packages/boltz-swaps/src/bridge/types.ts
```diff
@@ -29,7 +29,7 @@ import type {
 
 export type { BridgeTransaction };
 
-export type BridgeRoute = OftRoute;
+export type BridgeRoute<A extends string = string> = OftRoute<A>;
```
jackstar12: Shouldnt't it be the other way around that oft re-uses the BridgeRoute type?

packages/boltz-swaps/src/sdk.ts
michael1011: The properties are very different from swap type to type. They also have different features. I don't want to force it all into a singular type, throw `unknown` and `any` around just to hope for the best

jackstar12: Do we even need this in this PR already? The examples don't use it.

We can leave the low-level functionality in there, but we should eventually strive for implementors not having to care about what a reverse or submarine swap is. But completely different issue to be honest.

packages/boltz-swaps/src/client.ts
```diff
@@ -267,23 +257,21 @@ export const getPairs = async (options?: RequestInit): Promise<Pairs> => {
...
-    );
-    validateInvoiceForOffer(offer, res.invoice);
-
-    return res;
-};
+): Promise<{ invoice: string }> =>
+    fetcher<{ invoice: string }>("/v2/lightning/BTC/bolt12/fetch", {
+        offer,
+        amount: amountSat,
+    });
```
kiocos, line 271: this removed the validation `validateInvoiceForOffer(offer, res.invoice);` present in the function and didn't add the validation back on the webapp

kiocos (dismissed):

PR #1470, opened 2026-05-18
Embed styles and lock output

tompro:
This adds the following new features
- A light theme that currently only activates in embed mode and re-styles the swap form. It re-uses colors from default and pro. Can be activated via query param
- A `embedded` query param that reduces the full page view to only the swap form. Allows to show just the form in iFrame embeds
- A `lockOutput` query param. Currently this is only respected if the swap destination is an LN invoice. It disables the destination form element and hides the `paste invoice` input. It also prevents the destination (invoice) from being cleared when changing the sendAsset.

## Summary by CodeRabbit

* **New Features**
  * Embedded mode for Lightning invoices with URL-driven destination locking and automatic pre-fill of amounts
  * New light theme option

* **Improvements**
  * Inputs, controls and asset selectors disable or hide when destination is locked to prevent edits
  * Invoice/amounts preserved when destination is locked (no unintended resets)
  * Layout adapts when embedded; embedded branding relocated
  * Icon/filter and asset selection styling refined

[Review Change Stack](https://app.coderabbit.ai/change-stack/BoltzExchange/boltz-web-app/pull/1470?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/Asset.tsx
+++ b/src/components/Asset.tsx
@@ -30,6 +30,7 @@ const Asset = (props: {
     return (
         <button
             type="button"
+            disabled={props.disabled}
             class={`asset-wrap${
                 bitcoinOnly() || props.disabled ? " no-select" : ""
             }`}
--- a/src/components/Asset.tsx
+++ b/src/components/Asset.tsx
@@ -7,12 +7,18 @@ import { useCreateContext } from "../context/Create";
 import { useGlobalContext } from "../context/Global";
 import "../style/asset.scss";
 
-const Asset = (props: { side: Side; signal: Accessor<string> }) => {
+const Asset = (props: {
+    side: Side;
+    signal: Accessor<string>;
+    disabled?: boolean;
+}) => {
     const { bitcoinOnly, notify, t } = useGlobalContext();
 
     const openSelect = () => {
-        if (bitcoinOnly()) {
-            notify("error", t("bitcoin_only_warning"));
+        if (props.disabled || bitcoinOnly()) {
+            if (bitcoinOnly()) {
+                notify("error", t("bitcoin_only_warning"));
+            }
             return;
         }
         setAssetSelected(props.side);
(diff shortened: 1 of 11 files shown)


src/components/Asset.tsx
```diff
@@ -7,12 +7,18 @@ import { useCreateContext } from "../context/Create";
...
+    disabled?: boolean;
+}) => {
     const { bitcoinOnly, notify, t } = useGlobalContext();
 
     const openSelect = () => {
-        if (bitcoinOnly()) {
-            notify("error", t("bitcoin_only_warning"));
+        if (props.disabled || bitcoinOnly()) {
+            if (bitcoinOnly()) {
+                notify("error", t("bitcoin_only_warning"));
```
michael1011, line 20: Needs a test

src/components/AssetSelect.tsx
```diff
@@ -94,8 +95,10 @@ const SelectAsset = () => {
             return;
         }
 
-        // clear invoice every time asset changes
-        setInvoice("");
+        // clear invoice every time asset changes (unless destination is locked)
+        if (!destinationLocked()) {
+            setInvoice("");
+        }
```
michael1011, line 101: Also needs a test

src/context/Create.tsx
```diff
@@ -220,6 +224,35 @@ const handleUrlParams = (
...
+        try {
+            const sats = decodeInvoice(destination).satoshis;
+            if (sats > 0) {
+                setAmountChanged(Side.Receive);
+                setReceiveAmount(BigNumber(sats));
+                setQuoteLoading(true);
+                const sendAmt = await pair().calculateSendAmount(
+                    BigNumber(sats),
+                    minerFee(),
+                );
```
michael1011, line 242: I really don't like this function becoming async. Also feels like this could become racy when the pairs have not been fetched yet

src/context/Create.tsx
```diff
@@ -220,6 +224,35 @@ const handleUrlParams = (
...
+                setSendAmount(sendAmt);
+                if (
+                    embedded === "true" &&
+                    getUrlParam(UrlParam.LockOutput) === "true"
```
michael1011, line 246: Locking the output should be possible even when we are not embedding

michael1011 (changes requested):

michael1011:
Closed by https://github.com/BoltzExchange/boltz-web-app/pull/1498

PR #1473, opened 2026-05-18
refactor: replace logs copy with post to support

### LINKED ISSUE #957: Streamlined Log Submission
For support cases we currently have to ask users to go to Settings -> Logs -> Download and then upload the logfile to support chat or send via email. This is tedious, especially on mobile. I am thinking we should move from a file download (maybe keep it as an option) to a log submission via Chatwoot with a simple button press. A confirmation prompt should explain that this is a one-time sending of logs up to that moment, no continuous submission, no tracking.

michael1011:
Closes #957

### DIFF
--- a/src/utils/chatwoot.ts
+++ b/src/utils/chatwoot.ts
@@ -1,0 +1,69 @@
+import { config } from "../config";
+
+const chatwootAuthCookie = "cw_conversation";
+
+export const getChatwootToken = () =>
+    import.meta.env.VITE_CHATWOOT_TOKEN as string | undefined;
+
+export const isChatwootConfigured = () =>
+    !!getChatwootToken() && config.chatwootUrl !== undefined;
+
+const getCookie = (name: string) => {
+    const prefix = `${name}=`;
+    const cookie = document.cookie
+        .split(";")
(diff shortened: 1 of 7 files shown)


kiocos (commented):
Great stuff!

- [ ] Let's make the SettingsMenu close once the "share" button is clicked.
- [ ] For the sake of making it very obvious, I'd propose breaking the pattern a bit here and doing something like:

Image

src/utils/chatwoot.ts
```diff
@@ -0,0 +1,109 @@
...
+        : decodeURIComponent(cookie.substring(prefix.length));
+};
+
+export const logsFileName = "boltz-logs.json";
+
+export const formatLogsForChatwootAttachment = (
+    logs: Record<string, string[]>,
+) => new Blob([JSON.stringify(logs, null, 2)], { type: "application/json" });
+
+export const postMessageToChatwoot = async (content: string) => {
```
kiocos: this function should be deleted, never called

kiocos (changes requested):

kiocos (approved):

PR #1475, opened 2026-05-19
feat: make max button respect wallet balance

jackstar12:
## Summary by CodeRabbit

* **New Features**
  * Exact-input hop quoting via a new hop-input prop for lockups; hop input threaded into lockup flows
  * Swap limits: independent max-enable control and selectable resolved-maximum flow
  * Preserve and thread exact source amounts through bridge quoting, send flows and swap persistence

* **Bug Fixes**
  * More accurate native EVM lockup spendable-balance calculation

* **Tests**
  * Added tests for max-button behavior and lockup spendable-balance logic

[Review Change Stack](https://app.coderabbit.ai/change-stack/BoltzExchange/boltz-web-app/pull/1475?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/utils/swapCreator.ts
```diff
@@ -35,11 +35,15 @@ export type DexDetail = {
     // For post-swap hops: expected output amount from the DEX.
     // For pre-swap hops: expected input amount to the DEX.
     quoteAmount: number | string;
+
+    // For pre-swap hops, exact source amount to spend in the DEX.
+    sourceAmount?: string;
```
jackstar12, line 40: `quoteAmount` was populated previously but never used in the pre-swap dex path. So still introduced a seperate prop here to not mess with old swaps.

kiocos (approved):
tACK

kilrau (changes requested):
Can we get rid of the min button please? It's barely useful for anyone, always has the risk of slipping below our min and specifically with this PR it introes a very weird UX as it ingores my wallet balance, only max honors it and thus "max might be 0 or below min amount and then min fills  an amount > my balance."

https://github.com/user-attachments/assets/d949041c-a33f-4fa5-9f38-78a953c924f3

Otherwise works. Managed to empty my tBTC balance cleanly, also correctly recognized my balances on other chains I tested.

kilrau (approved):
LGTM

src/components/CreateButton.tsx
```diff
@@ -105,7 +107,48 @@ const buildBridgeDetail = (
...
+    const bridge = driver.getRoutePosition(route, position);
+    if (position !== SwapPosition.Pre || sourceAmount === undefined) {
+        return bridge;
+    }
+
+    return {
+        ...bridge,
+        sourceAmount: sourceAmount.toFixed(0),
+    };
+};
```
michael1011, line 119: Why are we doing this change?

src/components/CreateButton.tsx
```diff
@@ -105,7 +107,48 @@ const buildBridgeDetail = (
...
+    if (position !== SwapPosition.Pre || amountChanged !== Side.Send) {
+        return dex;
+    }
+
+    return {
+        ...dex,
+        sourceAmount: toDexAmount(
+            sendAmount.toNumber(),
+            hops[0].from,
+        ).toString(),
```
michael1011, line 150: Why?

src/components/LockupEvm.tsx
```diff
@@ -128,6 +154,15 @@ const getHopExecutionQuote = async (
...
 };
 
+const parseHopInputAmount = (amount: number | string | undefined) => {
+    if (amount === undefined) {
+        return undefined;
+    }
+    return typeof amount === "string"
+        ? BigInt(amount)
+        : BigInt(Math.round(amount));
+};
```
michael1011: How can that be either string, number or undefined?

src/components/LockupEvm.tsx
```diff
@@ -99,6 +100,31 @@ const getHopExecutionQuote = async (
...
+            });
+            throw new Error("could not get DEX quote for lockup hop");
+        }
+
+        const quote = quotes[0];
+        return {
+            quote,
+            amountIn: hopInputAmount,
+            targetLockupAmount: lockupAmount,
+        };
```
michael1011, line 125: Can we even do this for submarines? Just ignoring the slippage will cause swaps to fail

src/components/SendToBridge.tsx
```diff
@@ -121,28 +126,36 @@ const SendToBridge = (props: {
         const bridgeRoute = props.bridge;
         const quotedBridgeInstance =
             await bridgeDriver().getQuotedContract(bridgeRoute);
+
+        // we don't want to refetch a quote if we previously committed to
+        // a fixed source amount, e.g. for sweeping the wallet.
+        let sourceAmount = BigInt(bridgeRoute.sourceAmount ?? 0);
```
michael1011, line 132: That's nice if we don't want to, but we ***have*** to for submarines

src/pages/Create.tsx
```diff
@@ -561,6 +568,103 @@ const Create = () => {
         });
     };
 
+    const baseAssetAmountToInternal = (asset: string, amount: bigint) =>
+        getDecimals(asset).isErc20
+            ? BigNumber(amount.toString())
+            : BigNumber(assetAmountToSats(amount, asset).toString());
+
+    const getConnectedMaximum = async (currentPair: Pair) => {
```
michael1011: Completely untested

michael1011 (changes requested):

src/components/CreateButton.tsx
jackstar12: Because we have to commit to the source amount to send - we will get insufficient balance errors after swap creation otherwise if we have minor quote changes.

src/components/SendToBridge.tsx
jackstar12: This doesn't apply to submarines where you put in an invoice upfront - we don't commit to a send amount if the receive amount is the last one entered explicitly, which is always the case for submarine swaps in the current flow. Submarines where we have a commited source amount are implemented via https://github.com/BoltzExchange/boltz-web-app/pull/1492 properly. But this PR doesn't change how submarines are handled right now - we still handle slippage just like before if the *receive* amount is the last one entered

src/components/LockupEvm.tsx
jackstar12: https://github.com/BoltzExchange/boltz-web-app/pull/1475#discussion_r3333712764

michael1011 (approved):

PR #1492, opened 2026-05-21
feat: commitment swaps

### LINKED ISSUE #1208: EVM: "Max" button via commitment swaps
Problem: Users want to sweep their entire RBTC balance to e.g. Lightning. They have to manually calculate `balance` - `lockupGasCost` and set the invoice amount, which is very manual and error-prone and often leaves more dust than necessary.

Solution with commitment swaps:
- Client estimates gas for `lock()` (which has a known, bound gas cost)
- Client locks balance
- Client creates a swap for the actual locked amount and now precisely knows which invoice amount to set (or pull from lnurl/bolt12)
- Client submits the commitment signature

This gives users a seamless "Swap Max" experience. The only dust left is the rounding error in gas estimation, which is negligible and can't be avoided

jackstar12:


src/components/LockupEvm.tsx
```diff
@@ -668,19 +710,17 @@ const LockupEvm = (props: {
...
+                    preBridgeSourceAmount() ??
+                        (hasHopsBefore() && props.hops !== undefined
+                            ? (
+                                  await getHopExecutionQuote(
+                                      props.hops[0],
+                                      value(),
+                                      slippage(),
+                                  )
+                              ).amountIn
+                            : value()),
```
kiocos: `getHopExecutionQuote` returns
```
        return {
            quote,
            amountIn: hopInputAmount,
            targetLockupAmount: lockupAmount,
        };
```
without slippage and it uses the previous lockup quote.

Shouldn't we use the latest quote (from `quoteDexAmountIn`) and add the slippage?
Like `targetLockupAmount: calculateAmountOutMin(quotedOutput, slippage),`

src/status/CommitmentCreated.tsx
```diff
@@ -0,0 +1,438 @@
...
+    type CommitmentSwap,
+    createSubmarine,
+    getPreBridgeDetail,
+} from "../utils/swapCreator";
+import { validateInvoice, validateResponse } from "../utils/validation";
+
+const lockupAmountToInternal = (asset: string, amount: bigint) =>
+    getDecimals(asset).isErc20
+        ? BigNumber(amount.toString())
+        : BigNumber(assetAmountToSats(amount, asset).toString());
```
kiocos: This is duplicated code from `baseAssetAmountToInternal` from `pages/Create.tsx`

src/components/CreateButton.tsx
```diff
@@ -930,9 +969,73 @@ const CreateButton = () => {
...
+        const { gasAbstraction } = await getClaimAddress(
+            assetReceive,
+            assetSend,
+            signer,
+            onchainAddress,
+            getGasAbstractionSigner,
+            getGasToken(),
+        );
+
+        const dex =
```
kiocos:
```suggestion
        const dex: DexDetail | undefined =
```

src/utils/swapCreator.ts
```diff
@@ -227,6 +257,45 @@ const generatePreimage = ({
     return derivePreimageFromRescueKey(rescueFile, keyIndex, asset);
 };
 
+const createLocalSwapId = () => {
+    if (globalThis.crypto?.getRandomValues !== undefined) {
+        return hex.encode(globalThis.crypto.getRandomValues(new Uint8Array(4)));
+    }
```
kiocos:
```suggestion
    if (globalThis.crypto?.randomUUID !== undefined) {
        return globalThis.crypto.randomUUID();
    }
```
widely supported function, provides 122-bit ID
32-bit was prolly fine too, but it's a easy win

kiocos (changes requested):

kiocos:
Let's get it more consistent with our standard swap screen

image

image

Things I notice:
- [ ] We should use the same component for the "Amount" button, maybe place it under the invoice input?
- [ ] Should we display fees again on this page? If yes, can we make it more consistent with the Fees component used on the main page?
- [ ] When creating the commitment swap, I press the button "Create swap", which takes me to this page which has another "Create swap" button. Not sure how exactly we should name these, but the current naming feels off;
- [ ] If a user doesn't wish to move forward with the swap, would be nice to have some sort of "Refund" option;
- [ ] Lightning invoice validation is broken, `lightning:LNBC40720N1P4PN6CTPP56KHNL405XFJXVY90LJ0XC9AY9RVR0T22HYDEUWQ6MHAH09NQWLWQDPZ2DJKUEPQW3HJQNPDGF2YXGRPV3J8YETNWVCQZXRXQYP2XQRZJQDZWNDCVZP9MJ84P7NN7WJZ7CD7NPENT74MRY2AT7G4KA2YRUV6CCZZXEYQQ28QQQQQQQQQQQQQQQ9GQ2YSP5AJRP2DLC060NRDUJ459AG6CFV7S5ML49UVY9ANA32XDFNCVMCXES9QXPQYSGQP6QY50KVMHFPV08ZR9DV9AR6PW8S0SKKEKG4J288WJEC92WKP3U46072E89C4HTEVF70SHE5627HDX6PDNUKPUD674JNZMZT00C700SPSYR46M` is accepted during the main swap creation page, but results on an error on the commitment swap page;

@kilrau agree?

kilrau:
I played around with this and since commitment swaps (and "everything is a funding address" in future) are primarily relevant for Submarines, how about we don't actively help the user shooting himself in the foot and remove the Lightning destination input field when a. wallet is connected and b. the max amount button has been pressed and max amount successfully pulled from the wallet?

Reason: The `MAX` button is easy to discover. I press it. Then I see the Lightning invoice field, paste my invoice and... undid everything. We know from our 0 amount chain swaps that they are *very* hard to discover. That "Create Swap" button turning active without invoice is just very easy to miss and very hard to understand what it could mean.

I'd even slightly change the color of the sendAmount field and of the "MAX" button to signify "MAX AMOUNT MODE ON". Of course we let users click on the amount field or Max button again to deactivate it and change amounts around and all goes back to normal.

Tl;dr: a new "MAX AMOUNT" state. If this works, we can use a similar UX, like `FLEX` button for "FLEXIBLE AMOUNT" state to make our "Everything is a funding address" feature discoverable. It's the only way I see we can cover use cases like "consume entire UTXO/empty wallet" with "pay this specific Lightning invoice"

Wdyt? @jackstar12 @kiocos @michael1011 @jackstar12

kilrau (changes requested):
- [x] This should use the standard string including BOLT12 LNURL etc:
image

- [ ]  And yeah validation for bolt12 and lnurl fails which is annoying. it should use the exact same validtion and logic as in the swap box, also errors we do greyed out instead of ALARM red button all reusing the regular swap box states pretty please:
image

- [ ] And invoice amount off should show our usual error too:
Image

- [x] Why is that button suddenly green and not yellow? Also that refund button feels to prominent. It's new that we are allowing refunds before any permanent failed state so idk. I'm ok having it but not so much in the face. For starters switch order. Main action buttons always have lowest position in our swap flow.
Image

Other than that it worked and I managed to cleanly get my USDT balance to 0 🎉 

Image

src/status/CommitmentCreated.tsx
```diff
@@ -0,0 +1,593 @@
...
+                regularPairs: regularPairs(),
+            };
+        },
+        async ({
+            commitment: current,
+            pairs: currentPairs,
+            regularPairs,
+        }): Promise<CommitmentAmounts> => {
+            const provider = createAssetProvider(current.assetSend);
+            const receipt = await provider.waitForTransactionReceipt({
```
michael1011: This looks really brittle. One network hiccup and you get what looks like a failed swap

src/utils/rescue.ts
```diff
@@ -132,6 +132,7 @@ export const hasSwapTimedOut = (swap: SomeSwap, currentBlockHeight: number) => {
             (swap as ChainSwap).lockupDetails.timeoutBlockHeight,
         [SwapType.Reverse]: () => (swap as ReverseSwap).timeoutBlockHeight,
         [SwapType.Submarine]: () => (swap as SubmarineSwap).timeoutBlockHeight,
+        [SwapType.Commitment]: () => Number.MAX_SAFE_INTEGER,
```
michael1011: Not the case. Those have a unliteral refund timeout

src/utils/helper.ts
```diff
@@ -69,6 +69,17 @@ export const cropString = (str: string, maxLen = 40, subStrSize = 19) => {
     );
 };
 
+export const formatSwapId = (swap: { id: string; type: SwapType }) => {
+    if (swap.type !== SwapType.Commitment) {
+        return swap.id;
+    }
+
+    const legacyPrefix = "commitment-";
```
michael1011: What legacy?

michael1011 (changes requested):

PR #1507, opened 2026-06-01
feat: make max button respect wallet balance

michael1011:
Right base branch for https://github.com/BoltzExchange/boltz-web-app/pull/1475

## Summary by CodeRabbit

* **New Features**
  * “Set max” now derives the spendable maximum from the connected wallet (including gas-aware native EVM calculations).
  * Pre-route sending/locking can use an exact input source amount for quoting.

* **Improvements**
  * Bridge/swap sending now consistently tracks, passes, and persists the source amount end-to-end.
  * Hop-based lockups support an optional input override.
  * Swap Limits now shows only the maximum selector, with improved enable/loading behavior and a dedicated “max” action.

* **Localization**
  * Removed the “min” translation key.

* **Tests**
  * Updated and added coverage for max-only limits, connected maximum, and native EVM spendable balance.

### DIFF
--- a/packages/boltz-swaps/src/solana/lazy.ts
+++ b/packages/boltz-swaps/src/solana/lazy.ts
@@ -37,10 +37,8 @@ type SolanaOftModules = {
 export const solana: Loader<SolanaModules> = new Loader(
     "Solana",
     async (): Promise<SolanaModules> => {
-        const [web3, splToken] = await Promise.all([
-            import("@solana/web3.js"),
-            import("@solana/spl-token"),
-        ]);
+        const web3 = await import("@solana/web3.js");
+        const splToken = await import("@solana/spl-token");
 
         return { web3, splToken };
     },
@@ -49,9 +47,9 @@ export const solana: Loader<SolanaModules> = new Loader(
 export const solanaCctp: Loader<SolanaCctpModules> = new Loader(
     "Solana CCTP",
     async (): Promise<SolanaCctpModules> => {
-        const [solanaKit, web3, generated] = await Promise.all([
+        const web3 = await import("@solana/web3.js");
+        const [solanaKit, generated] = await Promise.all([
             import("@solana/kit"),
-            import("@solana/web3.js"),
             import("../generated/solana-cctp-token-messenger-minter/src/generated/index.ts"),
         ]);
 
@@ -66,6 +64,7 @@ export const solanaCctp: Loader<SolanaCctpModules> = new Loader(
 export const solanaOft: Loader<SolanaOftModules> = new Loader(
     "Solana OFT",
     async (): Promise<SolanaOftModules> => {
+        const web3 = await import("@solana/web3.js");
         const [
             lzSolanaUmi,
             mplToolbox,
@@ -74,7 +73,6 @@ export const solanaOft: Loader<SolanaOftModules> = new Loader(
             umiWalletAdapters,
             solanaKit,
             splToken,
-            web3,
             generated,
         ] = await Promise.all([
             import("@layerzerolabs/lz-solana-sdk-v2/umi"),
@@ -84,7 +82,6 @@ export const solanaOft: Loader<SolanaOftModules> = new Loader(
             import("@metaplex-foundation/umi-signer-wallet-adapters"),
             import("@solana/kit"),
             import("@solana/spl-token"),
-            import("@solana/web3.js"),
             import("../generated/solana-oft/src/generated/index.ts"),
         ]);
 
--- a/src/components/CreateButton.tsx
+++ b/src/components/CreateButton.tsx
@@ -580,6 +580,14 @@ const CreateButton = () => {
                     "hops" | "hopsPosition" | "sendAmount" | "receiveAmount"
                 >,
             ): Promise<string | undefined> => {
+                const sourceAmount =
+                    amountChanged() === Side.Send ? sendAmount() : undefined;
+                bridge =
+                    buildBridgeDetail(
+                        assetSend(),
+                        SwapPosition.Pre,
+                        sourceAmount,
+                    ) ?? buildBridgeDetail(assetReceive(), SwapPosition.Post);
                 dex =
                     creationData?.hops !== undefined &&
                     creationData?.hopsPosition !== undefined
@@ -588,16 +596,11 @@ const CreateButton = () => {
                               creationData.hopsPosition,
                               creationData.sendAmount,
                               creationData.receiveAmount,
+                              // If the bridge is involved, the source amount is already
+                              // persisted on the bridge and isn't involved in the DEX quote.
+                              bridge === undefined ? sourceAmount : undefined,
                           )
                         : undefined;
-                bridge =
-                    buildBridgeDetail(
-                        assetSend(),
-                        SwapPosition.Pre,
-                        amountChanged() === Side.Send
-                            ? sendAmount()
-                            : undefined,
-                    ) ?? buildBridgeDetail(assetReceive(), SwapPosition.Post);
 
                 const payload = buildSwapMetadataPayload({ dex, bridge });
                 const mnemonic = rescueFile()?.mnemonic;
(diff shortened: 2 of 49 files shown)


kilrau:
Follow up: https://github.com/BoltzExchange/boltz-web-app/issues/1517

packages/boltz-swaps/src/solana/lazy.ts
```diff
@@ -37,10 +37,8 @@ type SolanaOftModules = {
 export const solana: Loader<SolanaModules> = new Loader(
     "Solana",
     async (): Promise<SolanaModules> => {
-        const [web3, splToken] = await Promise.all([
-            import("@solana/web3.js"),
-            import("@solana/spl-token"),
-        ]);
+        const web3 = await import("@solana/web3.js");
+        const splToken = await import("@solana/spl-token");
```
michael1011, line 41: Why are those not a `Promise.all` anymore?

packages/boltz-swaps/src/solana/lazy.ts
```diff
@@ -49,9 +47,9 @@ export const solana: Loader<SolanaModules> = new Loader(
 export const solanaCctp: Loader<SolanaCctpModules> = new Loader(
     "Solana CCTP",
     async (): Promise<SolanaCctpModules> => {
-        const [solanaKit, web3, generated] = await Promise.all([
+        const web3 = await import("@solana/web3.js");
+        const [solanaKit, generated] = await Promise.all([
```
michael1011, line 51: Same here

packages/boltz-swaps/src/solana/lazy.ts
```diff
@@ -66,6 +64,7 @@ export const solanaCctp: Loader<SolanaCctpModules> = new Loader(
 export const solanaOft: Loader<SolanaOftModules> = new Loader(
     "Solana OFT",
     async (): Promise<SolanaOftModules> => {
+        const web3 = await import("@solana/web3.js");
```
michael1011, line 67: And here

src/utils/swapCreator.ts
```diff
@@ -260,6 +293,39 @@ const generatePreimage = ({
...
+    type: SwapType.Commitment,
+    gasAbstraction,
+    assetSend,
+    assetReceive,
+    date: new Date().getTime(),
+    version: OutputType.Taproot,
+    initialReceiveAsset,
+    sourceAsset,
+    sourceAmount: sourceAmount.toFixed(0),
+    lockupAmount: lockupAmount.toFixed(0),
```
michael1011: Do we have to set those to 0? Seems nasty to handle down the read when we don't know if a value is *actually* zero or just not initialized

src/utils/swapCreator.ts
```diff
@@ -260,6 +293,39 @@ const generatePreimage = ({
     return derivePreimageFromRescueKey(rescueFile, keyIndex, asset);
 };
 
+export const createLocalSwapId = () =>
+    Math.floor(Math.random() * 0x1_000000)
+        .toString(16)
+        .padStart(6, "0");
```
michael1011: Can we use something like this

```ts
Array.from(crypto.getRandomValues(new Uint8Array(8)), b => b.toString(16).padStart(2, "0")).join("")
```

To make them look more like our normal ids and avoid collisions

src/components/LockupEvm.tsx
```diff
@@ -100,6 +102,31 @@ const getHopExecutionQuote = async (
...
+                hopInputAmount: hopInputAmount.toString(),
+            });
+            throw new Error("could not get DEX quote for lockup hop");
+        }
+
+        const quote = quotes[0];
+        return {
+            quote,
+            amountIn: hopInputAmount,
+            targetLockupAmount: lockupAmount,
```
michael1011: This ignores the slippage entirely and uses the creation time value (which is 0 for commitments and will cause failed lockups for non commitments). `SwapExecutionWorker` already does handle this correctly with a fresh quote

src/components/SwapExecutionWorker.tsx
```diff
@@ -975,10 +980,14 @@ export const SwapExecutionWorker = () => {
...
                     currentSwap.commitmentLockupCallId,
                 );
+            const commitmentLockupDetails = await getCommitmentLockupDetails(
+                currentSwap.assetSend,
+            );
             await persistCommitmentLockupTransaction(
                 currentSwap.id,
                 commitmentLockupTxHash,
                 CommitmentLockupTransactionSource.Resumed,
+                Number(commitmentLockupDetails.timelock),
```
michael1011: This persists a potentially wrong value. We should store the timelock when actually sending the tx or read it from the sent tx

src/pages/Create.tsx
```diff
@@ -568,6 +584,49 @@ const Create = () => {
         });
     };
 
+    const setMaxAmount = async () => {
+        const selectedPair = pair();
+        let connectedMaximum: BigNumber | undefined;
+
+        try {
+            connectedMaximum = await getConnectedMaximum({
```
michael1011, line 594: This can take a while with eg slow RPCs or flaky internet. Let's disable the input while fetching

src/components/CreateButton.tsx
```diff
@@ -914,11 +904,77 @@ const CreateButton = () => {
...
+            onchainAddress,
+            getGasAbstractionSigner,
+            getGasToken(),
+        );
+
+        const dex = {
+            hops: creationData.hops,
+            position: SwapPosition.Pre,
+            quoteAmount: sendAmount().toFixed(0),
+            sourceAmount: sendAmount().toFixed(0),
```
michael1011: We should use `buildDexDetail` to not diverge. The amount handling here does already

src/components/CreateButton.tsx
```diff
@@ -506,20 +581,24 @@ const CreateButton = () => {
...
-                              quoteAmount:
-                                  creationData.hopsPosition ===
-                                  SwapPosition.Post
-                                      ? Number(creationData.receiveAmount)
-                                      : Number(creationData.sendAmount),
-                          }
+                        ? buildDexDetail(
+                              creationData.hops,
+                              creationData.hopsPosition,
+                              creationData.sendAmount,
```
kiocos, line 597: This is the wrong amount for pre-DEX send-side swaps. `creationData.sendAmount` is the amount after the DEX conversion, but `dex.sourceAmount` must be the amount the user originally sends into the DEX.

Creating a USDT0-ARB -> LBTC swap with this results in: 
Image
Instead of using the USDT amount I inserted for the DEX quote, it used the TBTC amount that would be sent to Boltz

kiocos (changes requested):

src/components/CreateButton.tsx
```diff
@@ -132,7 +132,7 @@ const buildDexDetail = (
     position: SwapPosition | undefined,
     sendAmount: BigNumber,
     receiveAmount: BigNumber,
-    amountChanged: Side,
+    sourceAmount?: BigNumber,
```
kiocos, line 135: `sourceAmount` not being passed here causes the lockup to switch from exact-input to exact-output.

Instead of telling the DEX "I want to spend exactly X USDT", we're telling “how much USDT do I need to get X TBTC?”.

That can either spend less than the selected max balance, leaving dust, or require more than the available max balance, causing the swap to fail.

src/components/CreateButton.tsx
```diff
@@ -506,20 +581,23 @@ const CreateButton = () => {
...
-                                  SwapPosition.Post
-                                      ? Number(creationData.receiveAmount)
-                                      : Number(creationData.sendAmount),
-                          }
+                        ? buildDexDetail(
+                              creationData.hops,
+                              creationData.hopsPosition,
+                              creationData.sendAmount,
+                              creationData.receiveAmount,
+                          )
```
kiocos, line 602: I believe this should be the fix for the previous comment
```suggestion
                        ? buildDexDetail(
                              creationData.hops,
                              creationData.hopsPosition,
                              creationData.sendAmount,
                              creationData.receiveAmount,
                              creationData.hopsPosition === SwapPosition.Pre &&
                              amountChanged() === Side.Send
                              ? sendAmount()
                              : undefined,
                          )
```

kiocos (changes requested):

kiocos (approved):

PR #1512, opened 2026-06-02
Post message notifier for swap completion events

tompro:
Given the app is embedded as an iFrame (embedded=true) and a parentOrigin=https://example.com is provided allows to send final swap status to the parent as a postMessage.

## Summary by CodeRabbit

* **New Features**
  * Support for a parentOrigin URL parameter and sending status notifications to the parent window when a swap reaches a final status.

* **Documentation**
  * Added docs describing parentOrigin, example notification payloads, and guidance for validating incoming postMessage events.

* **Tests**
  * Added unit and end-to-end tests verifying parent-window postMessage notification behavior.

### DIFF
--- a/src/utils/notifyParent.ts
+++ b/src/utils/notifyParent.ts
@@ -4,25 +4,28 @@ import { useGlobalContext } from "../context/Global";
 import { formatError } from "../utils/errors";
 
 export const useParentNotifier = () => {
-    const { embeddedMode, parentOrigin } = useGlobalContext();
+  const { embeddedMode, parentOrigin } = useGlobalContext();
 
-    const notifyParent = (message: unknown) => {
-        const origin = parentOrigin();
-        if (
-            !embeddedMode() ||
-            !origin ||
-            origin === "*" ||
-            window.parent === window
-        ) {
-            return;
-        }
+  const notifyParent = (
+    message: { type: string } & Record<string, unknown>,
+  ) => {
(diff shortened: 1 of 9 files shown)


src/utils/notifyParent.ts
```diff
@@ -0,0 +1,28 @@
+import log from "loglevel";
+
+import { useGlobalContext } from "../context/Global";
+import { formatError } from "../utils/errors";
+
+export const useParentNotifier = () => {
+    const { embeddedMode, parentOrigin } = useGlobalContext();
+
+    const notifyParent = (message: unknown) => {
```
michael1011:
```suggestion
    const notifyParent = (message: { type: string} & Record<string, unknown>) => {
```
Or something along the lines so that we always make sure to emit the `type` property

src/utils/notifyParent.ts
```diff
@@ -0,0 +1,28 @@
...
+        const origin = parentOrigin();
+        if (
+            !embeddedMode() ||
+            !origin ||
+            origin === "*" ||
+            window.parent === window
```
michael1011, line 18: `window.parent === window` needs a comment that this check tells us that we are not an iframe

michael1011 (commented):
Some e2e tests with playwright would be nice but not a blocker if those aren't easy to do and not worth the effort

michael1011 (approved):

PR #1518, opened 2026-06-04
feat: log currency denomination

### LINKED ISSUE #1403: Add currencies to amounts in logs
Currently web app logs show something like:
```
 "2026-04-29T12:05:26.725Z OFT received amount is less than expected {\"swapId\":\"sWaPiD\",\"guid\":\"0xABC\",\"amountReceived\":\"2572605\",\"amountExpected\":\"33480000000000\"}",
```

which is very hard to read, even for AI, as it involves metadata that is not in the logs: `2572605` is `2.572605 USDT`, `33480000000000` is `3348 TBTC sats`. Also, inferred from the fact that web app was expecting a different currency, we know the DEX hop from USDT -> TBTC didn't happen which is just infinitely easier if the currencies are added. Now that we have more than Bitcoin.

kilrau:
Closes https://github.com/BoltzExchange/boltz-web-app/issues/1403

## Summary by CodeRabbit

* **Chores**
  * Improved logging readability by adding denomination-aware formatting helpers and applying them across bridge sends, swap execution, lockups, quoting, transaction/claim flows and related UI/debug logs so amounts display as sats, decimals and symbols.

* **Tests**
  * Added tests validating the new formatting utilities across multiple assets and input types (string/BigInt) with fallback behavior to ensure consistent, readable log output.

### DIFF
--- a/src/components/SendToBridge.tsx
+++ b/src/components/SendToBridge.tsx
@@ -47,6 +47,10 @@ import { usePayContext } from "../context/Pay";
 import { type Signer, useWeb3Signer } from "../context/Web3";
 import type { DictKey } from "../i18n/i18n";
 import WalletConnectProvider from "../utils/WalletConnectProvider";
+import {
+    formatAssetAmountForLog,
+    formatNativeAmountForLog,
+} from "../utils/denomination";
 import { type BridgeDetail, GasAbstractionType } from "../utils/swapCreator";
 import ApproveErc20 from "./ApproveErc20";
 import ApproveTrc20 from "./ApproveTrc20";
@@ -356,15 +360,24 @@ const SendToBridge = (props: {
 
         log.info(`${logLabel} signer token balance check`, {
             asset: props.bridge.sourceAsset,
-            balance: balance.toString(),
-            requiredAmount: requiredTokenAmount.toString(),
+            balance: formatAssetAmountForLog(balance, props.bridge.sourceAsset),
+            requiredAmount: formatAssetAmountForLog(
+                requiredTokenAmount,
+                props.bridge.sourceAsset,
+            ),
             sufficient: hasEnoughTokenBalance,
         });
         log.info(`${logLabel} signer native balance check`, {
             asset: props.bridge.sourceAsset,
             destinationAsset: props.bridge.destinationAsset,
-            nativeBalance: nativeBalance.toString(),
-            requiredMsgFee: requiredNativeBalance.toString(),
+            nativeBalance: formatNativeAmountForLog(
+                nativeBalance,
+                props.bridge.sourceAsset,
+            ),
+            requiredMsgFee: formatNativeAmountForLog(
+                requiredNativeBalance,
+                props.bridge.sourceAsset,
+            ),
             sufficient: hasEnoughMsgFee,
         });
 
@@ -447,6 +460,7 @@ const SendToBridge = (props: {
         return {
             directSendTarget,
             recipient,
+            tokenAmount,
             sendParam,
             msgFee,
             hasEnoughTokenBalance,
@@ -651,8 +665,14 @@ const SendToBridge = (props: {
             sourceAsset: props.bridge.sourceAsset,
             destinationAsset: props.bridge.destinationAsset,
             recipient,
-            amount: tokenAmount.toString(),
-            nativeFee: msgFee[0].toString(),
+            amount: formatAssetAmountForLog(
+                tokenAmount,
+                props.bridge.sourceAsset,
+            ),
+            nativeFee: formatNativeAmountForLog(
+                msgFee[0],
+                props.bridge.sourceAsset,
+            ),
             lzTokenFee: msgFee[1].toString(),
         });
 
@@ -706,15 +726,22 @@ const SendToBridge = (props: {
             props.bridge,
             WalletConnectProvider.getTronProvider(),
         );
-        const { sendParam, msgFee } = await quoteBridgeSendState(recipient);
+        const { tokenAmount, sendParam, msgFee } =
+            await quoteBridgeSendState(recipient);
 
         log.debug("Quoted bridge send", {
             swapId: props.swapId,
             sourceAsset: props.bridge.sourceAsset,
             destinationAsset: props.bridge.destinationAsset,
             recipient,
-            amount: props.amount.toString(),
-            nativeFee: msgFee[0].toString(),
+            amount: formatAssetAmountForLog(
+                tokenAmount,
+                props.bridge.sourceAsset,
+            ),
+            nativeFee: formatNativeAmountForLog(
+                msgFee[0],
+                props.bridge.sourceAsset,
+            ),
             lzTokenFee: msgFee[1].toString(),
         });
 
@@ -735,6 +762,7 @@ const SendToBridge = (props: {
         const {
             directSendTarget,
             recipient,
+            tokenAmount,
             sendParam,
             msgFee,
             signerAddress,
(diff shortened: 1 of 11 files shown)


src/utils/denomination.ts
```diff
@@ -144,6 +150,31 @@ export const calculateDigits = (
...
 };
 
+/**
+ * Formats a raw on-chain amount (the asset's smallest unit: ERC20 base units,
+ * wei, or satoshis) into a readable log string with its currency symbol — e.g.
+ * `2572605` USDT becomes `"2.572605 USDT"`, `1016` BTC becomes `"1016 sats"`.
+ * BTC/L-BTC/Lightning render as sats; never throws (falls back to the raw value).
+ */
+export const formatAssetAmountForLog = (
+    amountLD: bigint | number | string,
```
michael1011: Why is this called `amountLD`

src/utils/denomination.ts
```diff
@@ -144,6 +150,31 @@ export const calculateDigits = (
     return digits;
 };
 
+/**
+ * Formats a raw on-chain amount (the asset's smallest unit: ERC20 base units,
+ * wei, or satoshis) into a readable log string with its currency symbol — e.g.
+ * `2572605` USDT becomes `"2.572605 USDT"`, `1016` BTC becomes `"1016 sats"`.
+ * BTC/L-BTC/Lightning render as sats; never throws (falls back to the raw value).
```
michael1011:
```suggestion
 * Formats a raw on-chain amount (the asset's smallest unit: ERC20 base units,
 * wei, or satoshis) into a readable log string with its currency symbol.
```

src/utils/denomination.ts
```diff
@@ -144,6 +150,31 @@ export const calculateDigits = (
...
+    amountLD: bigint | number | string,
+    asset: string,
+): string => {
+    try {
+        // BTC, L-BTC and Lightning amounts are already denominated in satoshis.
+        if (asset === LN || config.assets?.[asset]?.type === AssetKind.UTXO) {
+            return `${amountLD} sats`;
+        }
+        const sats = assetAmountToSats(BigInt(amountLD), asset);
+        const amount = satToBtc(BigNumber(sats.toString()));
```
michael1011: Why are we mix and matching denominations here?

michael1011 (commented):
There is unformatted amount prints left in `LockupEvm.tsx`, `src/components/CreateButton.tsx`, `src/utils/quoter`, `src/status/TransactionConfirmed.tsx`, `src/utils/Pair.ts`

src/components/SendToBridge.tsx
```diff
@@ -768,7 +778,10 @@ const SendToBridge = (props: {
             sourceAsset: props.bridge.sourceAsset,
             destinationAsset: props.bridge.destinationAsset,
             recipient,
-            amount: props.amount.toString(),
+            amount: formatAssetAmountForLog(
+                props.amount,
+                props.bridge.destinationAsset,
+            ),
```
kiocos, line 802: Since `props.amount is the amount Boltz needs to arrive on the destination chain (it was computed by the pair/DEX step on the far side)`, what we're printing here is actually the "receiveAmount", not the "sendAmount". We should print how much the user actually sends.

```suggestion
            amount: formatAssetAmountForLog(
                tokenAmount,
                props.bridge.destinationAsset,
            ),
```

To have `tokenAmount` here, first include it on the return of `refreshBridgeSendState`.

Then update the de-structure of `refreshBridgeSendState` at `sendBridgeFromEvm` with `tokenAmount`.

We use `tokenAmount` for Solana (`sendBridgeFromSolana`).

src/components/SendToBridge.tsx
```diff
@@ -713,7 +720,10 @@ const SendToBridge = (props: {
...
```
kiocos, line 740: Since `props.amount is the amount Boltz needs to arrive on the destination chain (it was computed by the pair/DEX step on the far side).`, what we're printing here is actually the "receiveAmount", not the "sendAmount". We should print how much the user actually sends.

```suggestion
            amount: formatAssetAmountForLog(
                tokenAmount,
                props.bridge.destinationAsset,
            ),
```

You can get `tokenAmount` from `quoteBridgeSendState` at line 716.

We use `tokenAmount` for Solana (`sendBridgeFromSolana`).

kiocos (changes requested):

src/utils/denomination.ts
```diff
@@ -144,6 +151,74 @@ export const calculateDigits = (
...
+        );
+        return `${formatted.toFixed()} ${getAssetDisplaySymbol(asset)}`;
+    } catch {
+        // Logging must never throw; fall back to the raw amount if the asset
+        // is unknown or cannot be denominated.
+        return `${amount.toString()} ${asset}`;
+    }
+};
+
+export const formatNativeAmountForLog = (
```
michael1011, line 184: Now we have three of those functions. Amazing. When do I use which?

tests/utils/denomination.spec.ts
```diff
@@ -175,4 +186,49 @@ describe("denomination utils", () => {
...
+        test("formats native amounts for logs", () => {
+            expect(formatNativeAmountForLog(10000000000000n, RBTC)).toEqual(
+                "0.00001 RBTC",
+            );
+        });
+
+        test.each`
+            amount            | asset    | expected
+            ${2572605n}       | ${USDT0} | ${"2.572605 USDT"}
+            ${3348n}          | ${TBTC}  | ${"3 348 sats"}
```
michael1011: Why do we format these swap amounts differently than everything else? That makes search in the logs horrible

src/utils/Pair.ts
```diff
@@ -978,17 +997,32 @@ export default class Pair {
...
-                messagingFeeCost: messagingFeeCost.toFixed(),
-                quotedAmountOut: quote.amountOut.toString(),
+                quotedBridgeAmount: formatSwapAmountForLog(
+                    quotedBridgeAmount,
+                    post.route.sourceAsset,
+                ),
+                messagingFee: formatNativeAmountForLog(
+                    bridgeMessagingFee,
+                    post.route.sourceAsset,
+                ),
```
michael1011: Messaging fee is in the native amount for OFT but not CCTP

michael1011 (changes requested):
Needs a rebase

There is still places to format the amounts missing in `src/pages/Create.tsx`, `TransactionLockupFailed.tsx` and `SendToBridge.tsx`

kilrau:
@coderabbitai full review

src/utils/denomination.ts
```diff
@@ -144,6 +151,77 @@ export const calculateDigits = (
...
+            ? value.div(BigNumber(10).pow(decimals))
+            : value;
+        return `${formatted.toFixed()} ${formatDenomination(
+            Denomination.Sat,
+            asset,
+        )}`;
+    } catch {
+        return `${amount.toString()} ${asset}`;
+    }
+};
```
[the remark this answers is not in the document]
kilrau: For now I decided to stick with the scale what the raw value already is. Not sure there is much of a value add for logs parsing/searches as gas amounts are usually not super interesting and also it just removes 10 decimals of on-chain precision. Happy to follow-up though if human reviewers share your concern.

kiocos (approved):
```
A past review comment flagged that formatting swap amounts differently makes searching logs harder. This is evident in the test cases:
- formatAssetAmountForLog(10000000000000n, RBTC) → "0.00001 RBTC" (on-chain wei)
- formatSwapAmountForLog(1016n, RBTC) → "1016 sats" (swap denomination)
```
May not be perfectly consistent, but is an improvement over what we had before. For most cases, it looks fine. I'd say this could be further improved later if we feel the need for it.

kilrau:
> ```
> A past review comment flagged that formatting swap amounts differently makes searching logs harder. This is evident in the test cases:
> - formatAssetAmountForLog(10000000000000n, RBTC) → "0.00001 RBTC" (on-chain wei)
> - formatSwapAmountForLog(1016n, RBTC) → "1016 sats" (swap denomination)
> ```
> 
> May not be perfectly consistent, but is an improvement over what we had before. For most cases, it looks fine. I'd say this could be further improved later if we feel the need for it.

Perfect? Definitely not. But one advantage is that it keeps 18 digit wei precision whenever that was the raw value. Also I just don't see the insane value or confusion that could be caused by it. I did this PR for myself so I'll iterate on it if I find it annoying.

michael1011 (approved):

PR #1520, opened 2026-06-05
feat: chain swap send support in sdk

michael1011:
## Summary by CodeRabbit

* **New Features**
  * SDK: swap status queries, chain-swap execute, route create/execute, cooperative UTXO claiming; new end-to-end example scripts (L‑BTC→TBTC, L‑BTC→USDC, Ark→USDC).

* **Refactor**
  * Centralized EVM/UTXO helpers, router-claim and slippage utilities, and unified transaction/claim flows for consistency.

* **Tests**
  * Extensive unit and regtest integration coverage for swaps, routing, EVM, UTXO, musig/cooperative flows.

* **Chores**
  * CI: added integration job; package version/exports and workspace catalog updated.


packages/boltz-swaps/integration/chainSwap.regtest.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+    BTC: getBitcoinAddress,
+    "L-BTC": getLiquidAddress,
+};
+
+const regtestUp = await isRegtestUp();
+if (!regtestUp) {
+    console.warn(
+        `[integration] Boltz regtest API unreachable at ${BOLTZ_API_URL} — skipping chain-swap integration tests.`,
+    );
+}
```
kiocos: When would this make sense? It could make ci pass even without running these tests

packages/boltz-swaps/integration/chainSwap.regtest.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+
+const mineChain: Record<UtxoAssetSym, () => Promise<string>> = {
+    BTC: generateBitcoinBlock,
+    "L-BTC": generateLiquidBlock,
+};
+
+const newClaimAddress: Record<UtxoAssetSym, () => Promise<string>> = {
+    BTC: getBitcoinAddress,
+    "L-BTC": getLiquidAddress,
+};
```
kiocos: This is not consistent with the pattern we have for our e2e tests, IMO we should invoke the functions directly instead of re-declaring a helper

packages/boltz-swaps/src/execute.ts
```diff
@@ -0,0 +1,193 @@
...
+        throw new Error(
+            `server lockup transaction for swap ${createdSwap.id} is unavailable`,
+        );
+    }
+
+    const claimDetails = createdSwap.claimDetails;
+    // The claim tx fee is funded by the gap between the gross lockup and the
+    // net receive amount, so default-to-gross leaves no fee budget. Callers
+    // claiming a UTXO destination pass the net (gross minus claim fee).
+    const receiveAmount = receiveAmountOverride ?? claimDetails.amount;
```
kiocos: Won't setting this to `claimDetails.amount` cause `const feeBudget = Math.floor(inputSum - receiveAmount);` to equal zero?

Also, shouldn't `receiveAmount` be mandatory?
There's a comment on this file saying `UTXO callers should always pass the net amount`.

packages/boltz-swaps/src/execute.ts
```diff
@@ -0,0 +1,193 @@
...
+        getSigner: () => signer,
+        gasAbstractionSigner: signer,
+        etherSwap,
+        erc20Swap,
+        sendTransaction: sendPopulatedTransaction,
+    });
+};
+
+const stripHexPrefix = (value: string): string =>
+    value.startsWith("0x") ? value.slice(2) : value;
```
kiocos: This is kinda duplicate from `stripHexPrefix` from `solana.ts` (this one handles uppercase too), which also duplicates from `events.ts`.
Let's move all the three to a single place and re-use it everywhere.

packages/boltz-swaps/src/config.ts
```diff
@@ -344,3 +363,12 @@ export const getReferralHeader = (): string | undefined =>
 
 export const isCooperativeDisabled = (): boolean =>
     getBoltzSwapsConfig().cooperativeDisabled === true;
+
+export const getConfiguredNetwork = (): "mainnet" | "testnet" | "regtest" =>
+    getBoltzSwapsConfig().network ?? "mainnet";
+
+export const getGasSponsorUrl = (): string =>
+    getBoltzSwapsConfig().gasSponsor ?? "https://sponsor.ccxp.space/";
```
kiocos:
```suggestion
export const getGasSponsorUrl = (): string =>
    getBoltzSwapsConfig().gasSponsor ?? MAINNET_API_DEFAULTS.gasSponsor;
```

kiocos (changes requested):

packages/boltz-swaps/src/routeExecute.ts
```diff
@@ -0,0 +1,442 @@
...
+    const [tradeQuote] = await quoteDexAmountIn(
+        dex.chain,
+        dex.tokenIn,
+        dex.tokenOut,
+        ctx.assetAmount,
+    );
+    const amountOutMin = calculateAmountOutMin(
+        BigInt(tradeQuote.quote),
+        ctx.slippage,
+    );
```
kiocos, line 336: `amoutOutMin` is calculated over a fresh quote, not over user's accepted quote. This could make slippage drift further than the maximum limit accepted by the user.

packages/boltz-swaps/src/execute.ts
```diff
@@ -0,0 +1,181 @@
...
+        claimAddress,
+        refundAddress: getAddress(claimDetails.refundAddress),
+        timeoutBlockHeight: claimDetails.timeoutBlockHeight,
+        destination,
+        getSigner: () => signer,
+        gasAbstractionSigner: signer,
+        etherSwap,
+        erc20Swap,
+        sendTransaction: sendPopulatedTransaction,
+    });
```
kiocos, line 82: depending on the asset, destination or claimAddress may or may not be used down the line, when sending the transaction. We should validate the expected variable for each asset. Correct me if I'm wrong, but I guess the requirements are:
```
ERC20 EVM:     destination may differ from claimAddress
Native EVM:   destination must be omitted or equal claimAddress
UTXO:         destination must be omitted or equal claimAddress
```
- For ERC20 EVM assets, `destination` may differ from `claimAddress` because the SDK can claim to `claimAddress` and then transfer the ERC20 to `destination`.
- For native EVM assets, the EtherSwap claim pays to `claimAddress` and there is no native forwarding step, so `destination` should be omitted or equal to `claimAddress`.
- For UTXO assets, the claim transaction is built with `claimAddress`; `destination` is not used, so it should also be omitted or equal to `claimAddress`.

kiocos (dismissed):

PR #1524, opened 2026-06-08
feat: patch dex swap with encrypted metadata and decrypt on external rescue

kiocos:
# PR 1 of 3
The three combined PRs will allow using the stored metadata to claim/refund to the user's original asset. 

1. https://github.com/BoltzExchange/boltz-web-app/pull/1524 data storage / retrieve
2. https://github.com/BoltzExchange/boltz-web-app/pull/1525 uses the returned metadata to enrich external rescue list (show the original pairs on the swaps list), also dedups results, preserves previous `main` behavior if metadata is not available / there is no match for it
3. https://github.com/BoltzExchange/boltz-web-app/pull/1549 final step: claim / refund the original asset via the external rescue flow

## Summary by CodeRabbit

* **New Features**
  * Swap creation and restoration now support encrypted swap route metadata, enabling more complete external rescue recovery.
  * Added the ability to update swap metadata after creation (including after lockup is persisted).

* **Bug Fixes**
  * Encrypted metadata is decrypted during rescue scanning and safely ignored on failure, preventing flow breakage.
  * Improved robustness when lockup transaction IDs are missing.
  * HTTP requests with parameters now respect an explicitly set HTTP method.


src/utils/swapMetadataPatch.ts
jackstar12, line 1: lets merge this with `swapMetadata.ts`

src/components/LockupEvm.tsx
```diff
@@ -576,6 +577,7 @@ const LockupTransaction = (props: {
                             `missing swap ${props.swapId} for lockup persistence`,
                         );
                     }
+                    await patchEncryptedSwapMetadata(updated, rescueFile());
```
jackstar12, line 584: What if we hook this into `modifySwap` instead of trying to remember calling it everywhere?

src/components/CreateButton.tsx
```diff
@@ -113,6 +117,46 @@ const buildBridgeDetail = (
     return driver.getRoutePosition(route, position);
 };
 
+export const buildEncryptedSwapMetadata = async ({
```
jackstar12: Dead code

src/pages/external-rescue/scan.ts
```diff
@@ -135,10 +139,40 @@ export const fetchPaginatedRestorableSwaps = async (
...
+
+            try {
+                const metadata = await decryptSwapMetadata(
+                    mnemonic,
+                    swap.metadata,
+                );
+                return {
+                    ...base,
+                    ...swapMetadataToLocalFields(metadata),
+                };
```
jackstar12, line 159: Lets merge decrypt + parse into one. Both fuctions are only ever used here. can still be split in a local helper for testing if you want

tests/utils/swapCreator.spec.ts
```diff
@@ -165,3 +187,75 @@ describe("getFinalAssetReceive", () => {
...
+    });
+    const newKeyNone: newKeyFn = vi.fn().mockResolvedValue(undefined);
+
+    beforeEach(() => {
+        createSubmarineMock.mockReset().mockResolvedValue({ id: "sub" });
+        createReverseMock.mockReset().mockResolvedValue({ id: "rev" });
+        createChainMock.mockReset().mockResolvedValue({ id: "chain" });
+    });
+
+    test("createSubmarine does not forward metadata", async () => {
```
jackstar12: These tests necessary?

src/utils/swapMetadata.ts
```diff
@@ -0,0 +1,213 @@
...
+export type SwapMetadataLocalFields = {
+    dex?: {
+        hops: EncodedHop[];
+        position: SwapPosition;
+        quoteAmount: number | string;
+    };
+    bridge?: BridgeDetail;
+    lockupTx?: string;
+    commitmentLockupTxHash?: string;
+};
```
jackstar12: Let's rather derive from the original swap types and not re-define

jackstar12 (dismissed):

src/utils/swapMetadata.ts
```diff
@@ -0,0 +1,213 @@
...
+
+// Converts a decrypted payload back into the local swap fields. Used during
+// restore to repopulate DEX/bridge routes the backend does not store.
+export const swapMetadataToLocalFields = (
+    payload: SwapMetadataPayload,
+): SwapMetadataLocalFields => {
+    const fields: SwapMetadataLocalFields = {};
+
+    if (payload.position !== undefined && payload.hops !== undefined) {
```
jackstar12: Maybe we wanna be more strict in the parsing to avoid unexpected behaviour if the metadata schema ever changes in the future?

src/components/LockupEvm.tsx
kiocos: we only need to call it one time per swap (once we have the broadcast tx id for dex swaps), so better to have it separate, otherwise this logic would run unnecessarily every time we run `modifySwap`

packages/boltz-swaps/tests/client.spec.ts
```diff
@@ -125,3 +127,35 @@ describe("getSwapStatuses", () => {
...
     });
 });
+
+describe("boltzClient swap metadata", () => {
+    beforeEach(() => {
+        fetcherMock.mockReset();
+        fetcherMock.mockResolvedValue({});
+    });
+
+    test("create requests do not forward metadata", async () => {
```
michael1011: Why aren't we doing that? Like this, all Arbitrum swaps without lockups still show up as TBTC

src/context/Pay.tsx
```diff
@@ -207,9 +209,16 @@ const PayProvider = (props: { children: JSX.Element }) => {
...
-                await modifySwap(swapId, (s) => {
+                if (lockupTxId === undefined) {
+                    return;
+                }
+
+                const updatedSwap = await modifySwap(swapId, (s) => {
                     s.lockupTx = lockupTxId;
                 });
+                if (updatedSwap !== null) {
+                    await patchEncryptedSwapMetadata(updatedSwap, rescueFile());
```
michael1011: Why are we updating here? This has nothing to do with lockups

src/utils/swapMetadata.ts
```diff
@@ -0,0 +1,309 @@
...
+    "lockupTx",
+    "commitmentLockupTxHash",
+]);
+const dexMetadataKeys = new Set(["hops", "position", "quoteAmount"]);
+const bridgeMetadataKeys = new Set([
+    "sourceAsset",
+    "destinationAsset",
+    "kind",
+    "position",
+]);
```
michael1011: I really want some type safety here. This could easily drift from our schema without noticing

src/utils/swapMetadata.ts
```diff
@@ -0,0 +1,309 @@
...
+        false,
+        ["deriveKey"],
+    );
+
+    return crypto.subtle.deriveKey(
+        {
+            name: "HKDF",
+            hash: "SHA-256",
+            salt: new Uint8Array(0),
+            info: new Uint8Array(0),
```
michael1011: Let's set the info to something like `swapMetadata/aes-gcm/v1`

michael1011 (changes requested):

michael1011 (approved):

PR #1526, opened 2026-06-09
feat: commitment swaps hover clear

jackstar12:


kilrau:
One NIT weird behavior where my lnurl/bolt12 "freezes" when I got max amount selected and try to "correct" my lnurl/bolt12:
[Screencast from 2026-06-22 18-32-47.webm](https://github.com/user-attachments/assets/4ae8ae8d-0c5e-499d-8225-cc20fa02a714)

Otherwise all these are addressed and played around with the different swap creation scenarios and deemed gut:
- [x] if invoice pasted and "max" button or manually changing USD amounts -> tool tip explaining why invoice was cleared
- [x] lnurl/bolt12 with manual usd (manual amount or max) -> commitment flow with invoice fetching after funding
- [x] when wallet empty -> max amount should fall back to max swap amount and not set to 0
- [x] change invoice/lnurl/bolt12 string when greyed out + mobile behavior of "Clear amounts" -> probably string change does the trick

PR #1528, opened 2026-06-10
feat: add unix timestamp to rescue file name

kiocos:
## Summary by CodeRabbit

* **New Features**
  * Backup downloads now use date-based filenames (timestamped) instead of a fixed name, making it easier to identify and manage multiple backup versions.
* **Behavior**
  * Download behavior and error handling remain unchanged; backups still validate presence before downloading.

### DIFF
--- a/src/utils/backup.ts
+++ b/src/utils/backup.ts
@@ -3,7 +3,10 @@ import type { Accessor } from "solid-js";
 import { downloadJson } from "./download";
 import type { RescueFile } from "./rescueFile";
 
-export const rescueFileName = "boltz-rescue-key-DO-NOT-DELETE";
+const rescueFileNamePrefix = "boltz-rescue-key-DO-NOT-DELETE";
+
+export const getRescueFileName = (date = new Date()) =>
+    `${rescueFileNamePrefix}-${Math.floor(date.getTime() / 1000)}`;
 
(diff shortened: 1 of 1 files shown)


kilrau (changes requested):
`boltz-rescue-key-DO-NOT-DELETE-06-11-26-0724PM` this is a a very specific and opinionated formatting. Also, as most users only ever download one resue file, that timestamp info might confuse (do I need to get a new one for every swap?).

So I'd vouch for unix timestamps just like we use them for history exports. It has the nice advantage of not bothering regular users, it feels like a random string of numbers, an id and for privacy browser power users unix time stamps allow rescue key files to be easily sorted in the filesystem. And they still can find out the exact human readable time if they want to. It's a power user feature after all.

kiocos:
> `boltz-rescue-key-DO-NOT-DELETE-06-11-26-0724PM` this is a a very specific and opinionated formatting. Also, as most users only ever download one resue file, that timestamp info might confuse (do I need to get a new one for every swap?).
> 
> So I'd vouch for unix timestamps just like we use them for history exports. It has the nice advantage of not bothering regular users, it feels like a random string of numbers, an id and for privacy browser power users unix time stamps allow rescue key files to be easily sorted in the filesystem. And they still can find out the exact human readable time if they want to. It's a power user feature after all.

done!

kilrau (approved):
tACK

PR #1529, opened 2026-06-11
fix: stranded funds from dex quote shortfall

### LINKED ISSUE #1446: Pre-bridge DEX quote shortfall silently strands bridged funds
## What happens

In the bridge → DEX → lockup path, after the bridge receive event is observed, `SwapExecutionWorker` fetches a DEX quote for the bridged amount. If the quote's `amountOut` is below the required lockup amount, the worker logs `"Bridge received amount is less than expected"` and returns early. The bridged token sits on the destination chain and the swap stops progressing.

## Where

`src/components/SwapExecutionWorker.tsx:1008-1016`

```ts
if (quote.trade.amountOut < expectedAmount) {
    log.warn("Bridge received amount is less than expected", { … });
    return;
}
```

## Issues

1. **No user-facing signal.** Only a `log.warn`; the UI doesn't surface that the swap is blocked or that funds need to be recovered.
2. **Misleading log fields.** `amountReceived` is in the bridge destination token's decimals while `amountExpected` is in `assetSend`'s decimals — they aren't directly comparable. The actual check uses `quote.trade.amountOut` (in `assetSend` units), which isn't logged in this warning. The preceding `log.debug("Swap execution fetched pre-bridge DEX quote", …)` has it, but only at debug level.
3. **No retry / recovery path.** A transient slippage or liquidity dip permanently parks the swap until the user manually recovers.

## Suggestions

- Include `quotedAmountOut` and `slippage` in the warning so it's self-contained.
- Surface a swap-level error state ("DEX quote insufficient — recover bridged funds") in the UI.
- Consider polling the quote a few times before giving up, to ride out transient DEX conditions.

kiocos:
Closes #1446

**How to easily test it:**
1. Create a USDT/USDC swap which includes a pre-bridge;
2. After you created the swap, update the function `fetchDexQuote` from `src/utils/quoter.ts`, so it does this early return (you can adjust the percentage to decide how much you want the quote to drift):
```
export const fetchDexQuote = async (
    hop: Hop,
    amountIn: bigint,
    getGasToken?: boolean,
    gasTokenAmount?: bigint,
): Promise<ClaimQuote> => {
     const debugTrade = await fetchQuote(Direction.In, hop, amountIn);
    return {
        trade: {
            ...debugTrade,
            amountOut: (debugTrade.amountOut * 60n) / 100n,
        },
    };
```
3. Press the "Send" button to bridge the asset;
4. Once the bridge is completed, the new error screen will show up, allowing the user to refund or retry the quote.

- if quote drifts within the slippage, swap continues normally
- if quote drifts outside the slippage on chain swaps, I show the renegotiation flow (Accept / refund buttons)
- if quote drifts outside the slippage on submarine swaps, I just show the "Retry quote" + "Refund" buttons
- If quote drifts below the absolute minimum for the pair, I just show the "Retry quote" + "Refund" buttons

Refunds are always sent back to the original network

## Summary by CodeRabbit

* **New Features**
  * Added a pre-bridge DEX quote “blocked” recovery flow with configurable retry attempts and a guided recover action.
* **UI Enhancements**
  * Updated the Pay page to display the blocked-quote recovery screen when pre-bridge recovery is present, and refined loading/refund UI details.
* **Localization**
  * Added new multilingual status messages for “refunded bridge pending” and blocked-quote states.
* **Bug Fixes**
  * Improved quote sufficiency checks to use slippage-adjusted thresholds and refined pre-bridge eligibility for retrying states.
  * Enhanced reverse-bridge refund follow-up call construction for pre-bridge scenarios.
* **Tests**
  * Added/expanded coverage for retry, recovery, and slippage boundary behavior.

### DIFF
--- a/src/components/RefundButton.tsx
+++ b/src/components/RefundButton.tsx
@@ -58,7 +58,12 @@ import {
 import { type AlchemyCall, toAlchemyCall } from "../alchemy/Alchemy";
 import RefundEta from "../components/RefundEta";
 import { config } from "../config";
-import { type AssetType, getKindForAsset, isEvmAsset } from "../consts/Assets";
+import {
+    type AssetType,
+    getKindForAsset,
+    getTokenAddress,
+    isEvmAsset,
+} from "../consts/Assets";
 import { type deriveKeyFn, useGlobalContext } from "../context/Global";
 import { usePayContext } from "../context/Pay";
 import { type Signer, useWeb3Signer } from "../context/Web3";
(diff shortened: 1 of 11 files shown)


src/components/SwapExecutionWorker.tsx
```diff
@@ -251,6 +259,163 @@ export const SwapExecutionWorker = () => {
...
+            // threshold. When the lockup already clears the chain swap
+            // renegotiation minimum, the backend can renegotiate the shortfall,
+            // so lock immediately instead of burning the remaining attempts.
+            const lockupAmount = calculateAmountOutMin(
+                quote.trade.amountOut,
+                slippage(),
+            );
+            if (
+                swap.type === SwapType.Chain &&
+                (await preBridgeLockupClearsRenegotiationMinimum(
```
jackstar12: This function name is getting a bit crazy for the amount of work it does. lets either inline completely or just have it as a local fn

src/utils/swapCreator.ts
```diff
@@ -55,6 +55,30 @@ export type PendingBridgeSendCallbacks = {
     persist: (pending: PendingBridgeSend) => Promise<void>;
 };
 
+export type PreBridgeRecoveryCall = {
```
jackstar12: Don't we already have a type for basic call data somewhere?

src/status/PreBridgeDexQuoteBlocked.tsx
```diff
@@ -0,0 +1,216 @@
...
+                amount: BigInt(currentRecovery.amount),
+                slippage: slippage(),
+                dexDetails: currentSwap.dex,
+                bridge: currentSwap.bridge,
+            },
+        );
+        const calls: AlchemyCall[] = [
+            ...(currentRecovery.receiveCall === undefined
+                ? []
+                : [toAlchemyCall(currentRecovery.receiveCall)]),
```
jackstar12: Lets invert this thing where we initialize calls as empty, optionally append the recieveCall and then append all the other calls

src/status/PreBridgeDexQuoteBlocked.tsx
```diff
@@ -0,0 +1,216 @@
...
+        const txHash = await sendPopulatedTransaction(
+            GasAbstractionType.Signer,
+            gasAbstractionSigner,
+            calls,
+        );
+        log.info(
+            `Refunded bridged funds for swap ${currentSwap.id} in ${txHash}`,
+        );
+
+        await persistRecovery({
```
jackstar12, line 128: cant we just delete the whole local state here

src/components/SwapExecutionWorker.tsx
kiocos: extracted part of it into a smaller helper + renamed the functions

src/status/PreBridgeDexQuoteBlocked.tsx
kiocos: no, we gotta keep track of the swap latest state (user might come back to the swap page later)

src/status/PreBridgeDexQuoteBlocked.tsx
```diff
@@ -0,0 +1,216 @@
...
+                                                <BlockExplorer
+                                                    asset={
+                                                        currentRecovery.asset
+                                                    }
+                                                    kind={
+                                                        BlockExplorerTargetKind.Tx
+                                                    }
+                                                    id={txHash}
+                                                    typeLabel="refund_tx"
+                                                />
```
jackstar12, line 164: Let's make this a proper link to the bridge explorer

src/components/RefundButton.tsx
```diff
@@ -360,6 +322,128 @@ const buildRefundFollowUpCalls = async (
...
+        sourceToken: refundData.tokenAddress,
+        amount: refundData.amount,
+        quoteChain,
+        refundAddress: refundData.refundAddress,
+        slippage,
+        trade: { desiredToken },
+    });
+};
+
+export const buildPreBridgeReverseBridgeRefundCalls = async ({
```
jackstar12: This function name is also getting out of hand. In fact, let's not have this here at all and simply inline those few calls in the pre bridge component since its just a bit of validation and calling the existing function. It's also not used anywhere else either

src/utils/swapCreator.ts
```diff
@@ -102,6 +121,10 @@ export type SwapBase = {
 
     // Bridge routes for bridging before lockup or after claim.
     bridge?: BridgeDetail;
+
+    // Local execution state for client-side swap steps that are not backend
+    // statuses.
+    execution?: SwapExecutionState;
```
jackstar12: Maybe we could tie this into the bridge state since its only relevant for bridged swaps anyways instead of a new execution state?

jackstar12 (dismissed):

michael1011 (approved):

PR #1536, opened 2026-06-20
fix: swap write race

michael1011:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DEX quote failure handling with explicit guards and deterministic error messages when no quote is returned (including zero-amount requests).

* **Refactor**
  * Switched swap state updates to a safer serialized read-modify-write flow, reducing inconsistent or conflicting swap updates during concurrent actions.

* **Tests**
  * Updated mocks and added coverage for the new swap update mechanism and the DEX quote guard behavior.

### DIFF
--- a/src/components/SwapExecutionWorker.tsx
+++ b/src/components/SwapExecutionWorker.tsx
@@ -64,6 +64,7 @@ import { swapStatusPending } from "../consts/SwapStatus";
 import { useGlobalContext } from "../context/Global";
 import { usePayContext } from "../context/Pay";
 import { useWeb3Signer } from "../context/Web3";
+import { useModifySwap } from "../hooks/useModifySwap";
 import { formatAssetAmountForLog } from "../utils/denomination";
 import { sendPopulatedTransaction } from "../utils/evmTransaction";
 import { decodeInvoice } from "../utils/invoice";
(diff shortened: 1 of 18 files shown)


src/components/SwapExecutionWorker.tsx
```diff
@@ -218,36 +218,41 @@ const isTaskRelevant = (
...
+    ): Promise<SomeSwap | null> => {
+        const updated = await modifySwap(id, mutator);
 
-        if (swap()?.id === updatedSwap.id) {
-            setSwap(updatedSwap);
+        if (updated !== null && swap()?.id === updated.id) {
+            setSwap(updated);
         }
+
+        return updated;
```
kiocos: The logic of this function is duplicated at:
1. `Pay.tsx:255–257`
2. `Pay.tsx:312–317`, 
3. `TransactionConfirmed.tsx:898–900`
4. `TransactionConfirmed.tsx:1176–1178`
5. `TransactionLockupFailed.tsx:269–271`

Would be better to have a centralized function on `PayContext`.

I'd also take the opportunity to fix the naming, which currently feels backwards.
At the moment:
`modifySwap` is the one that actually modifies the storage
`modifyStoredSwap` is more about updating the UI

kiocos (changes requested):
There are direct `setSwapStorage` and `updateSwapStatus` calls in the code that write to storage directly and bypass the lock. We should ensure consistent call of the function with the lock.

kiocos (approved):

PR #1542, opened 2026-06-28
feat: WebSocket ping/pong to prevent stalling

michael1011:
## Summary by CodeRabbit

* **New Features**
  * Added WebSocket keep-alive checks to detect stalled connections and reconnect automatically.
  * Re-sends active swap subscriptions after a connection is restored.

* **Bug Fixes**
  * Stops heartbeat pings when the socket closes or the component unmounts.
  * Prevents outdated connections from continuing to schedule reconnects.
  * Improves handling when messages stop arriving, reducing silent connection freezes.

### DIFF
--- a/src/components/SwapChecker.tsx
+++ b/src/components/SwapChecker.tsx
@@ -25,11 +25,14 @@ type SwapStatus = {
 };
 
 const reconnectInterval = 5_000;
+const pingInterval = 15_000;
 
-class BoltzWebSocket {
+export class BoltzWebSocket {
     private ws?: WebSocket;
     private reconnectTimeout?: ReturnType<typeof setTimeout>;
     private isClosed: boolean = false;
+    private pingTimer?: ReturnType<typeof setInterval>;
+    private awaitingPong: boolean = false;
 
     constructor(
         private readonly url: string,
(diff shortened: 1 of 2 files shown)


src/components/SwapChecker.tsx
```diff
@@ -43,7 +46,7 @@ class BoltzWebSocket {
 
     public connect = () => {
         log.debug("Opening WebSocket");
-        void this.openWebSocket(`${this.url}/v2/ws`);
+        void this.openWebSocket(`${this.url}/v2/ws`).catch(() => {});
```
jackstar12, line 49: Why swallow that error?

src/components/SwapChecker.tsx
```diff
@@ -88,7 +92,15 @@ class BoltzWebSocket {
     private openWebSocket = (url: string) => {
         this.isClosed = false;
         clearTimeout(this.reconnectTimeout);
-        this.ws?.close();
+        this.stopPinging();
+
+        if (this.ws !== undefined) {
```
jackstar12, line 97: Why cant we stop pinging and close like before?

src/components/SwapChecker.tsx
michael1011: Was always ignored. This just hides the unhandled rejection from the console

src/components/SwapChecker.tsx
michael1011: To cleanly shut down the socket. We didn't do that before

PR #1543, opened 2026-06-29
feat: WebSocket support in SDK

michael1011:
Depends on https://github.com/BoltzExchange/boltz-backend/pull/1451

## Summary by CodeRabbit

* **New Features**
  * Added real-time swap status streaming and callback subscriptions via `statusSource`, including async `watch` and batched status lookup (`swap.statuses`).
  * Introduced a configurable `statusSource` with automatic WebSocket-to-REST fallback, terminal-state stopping options, and new public `statusSource` package entry.
* **Bug Fixes**
  * Updated swap examples and SwapChecker to use streamed status updates instead of fixed-interval polling.
* **Tests**
  * Added/expanded unit and integration coverage for batch queries, streaming semantics, polling/WebSocket fallback behavior, and updated SwapChecker logic.

### DIFF
--- a/packages/boltz-swaps/src/statusSource/polling.ts
+++ b/packages/boltz-swaps/src/statusSource/polling.ts
@@ -1,0 +1,211 @@
+import { getSwapStatus, getSwapStatuses } from "../client.ts";
+import { safeEmit, safeEmitError } from "./emit.ts";
+import type {
+    StatusErrorHandler,
+    StatusSource,
+    StatusUpdateHandler,
+    SwapUpdate,
+    Unsubscribe,
+} from "./types.ts";
+
+export type PollingStatusSourceOptions = {
+    // Delay between fetches in ms. Default 5000.
+    intervalMs?: number;
+};
+
+type PollingEntry = {
+    handlers: Set<StatusUpdateHandler>;
+    errorHandlers: Set<StatusErrorHandler>;
+    lastUpdate?: SwapUpdate;
+    lastSignature?: string;
+};
+
+// REST polling StatusSource: one shared loop bulk-fetches all tracked ids per
+// interval (GET /v2/swap/status) and fans each out to its subscribers, emitting
+// only when an id's status changes.
+export const createPollingStatusSource = (
+    options: PollingStatusSourceOptions = {},
+): StatusSource => {
+    const intervalMs = options.intervalMs ?? 5_000;
+    const entries = new Map<string, PollingEntry>();
+
+    let intervalTimer: ReturnType<typeof setTimeout> | undefined;
+    let immediateTimer: ReturnType<typeof setTimeout> | undefined;
+    let polling = false;
+    let repollRequested = false;
+    let closed = false;
+
+    const deliver = (id: string, update: SwapUpdate): void => {
+        const entry = entries.get(id);
+        if (entry === undefined) {
+            return;
+        }
+        entry.lastUpdate = update;
+        const signature = JSON.stringify(update);
+        if (entry.lastSignature === signature) {
+            // Unchanged since the last delivery: skip, matching the WebSocket
+            // source, which only emits when the backend pushes a change.
+            return;
+        }
+        entry.lastSignature = signature;
+        for (const handler of entry.handlers) {
+            safeEmit(handler, update);
+        }
+    };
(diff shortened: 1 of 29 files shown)


michael1011:
@coderabbitai review

packages/boltz-swaps/src/statusSource/polling.ts
```diff
@@ -0,0 +1,232 @@
...
+    let repollRequested = false;
+    let closed = false;
+
+    const deliver = (id: string, update: SwapUpdate): void => {
+        const entry = entries.get(id);
+        if (entry === undefined) {
+            return;
+        }
+        entry.lastUpdate = update;
+        const signature = signatureOf(update);
```
jackstar12: A JSON.stringify would also do the trick and be less fragile here

packages/boltz-swaps/src/statusSource/polling.ts
```diff
@@ -0,0 +1,232 @@
...
+                return () => {};
+            }
+            let entry = entries.get(id);
+            const isNew = entry === undefined;
+            if (entry === undefined) {
+                entry = { handlers: new Set(), errorHandlers: new Set() };
+                entries.set(id, entry);
+            }
+            const handlerIsNew = !entry.handlers.has(onUpdate);
+            entry.handlers.add(onUpdate);
```
jackstar12: shouldnt we gate this behind `handlerIsNew`?

packages/boltz-swaps/src/statusSource/types.ts
```diff
@@ -0,0 +1,23 @@
+export type SwapUpdate = {
+    id: string;
+    status: string;
+    failureReason?: string;
+    zeroConfRejected?: boolean;
+    transaction?: { id: string; hex?: string };
+};
+
+export type StatusUpdateHandler = (update: SwapUpdate) => void;
+export type StatusErrorHandler = (error: unknown, id: string) => void;
```
jackstar12: Let's do an error type here - we have a defined error schema dont we?

packages/boltz-swaps/src/statusSource/polling.ts
```diff
@@ -0,0 +1,232 @@
...
+const signatureOf = (update: SwapUpdate): string =>
+    [
+        update.status,
+        update.failureReason ?? "",
+        update.zeroConfRejected ?? "",
+        update.transaction?.id ?? "",
+        update.transaction?.hex ?? "",
+    ].join("|");
+
+const safeEmit = (handler: StatusUpdateHandler, update: SwapUpdate): void => {
```
jackstar12: This is duplicated with the `websocket.ts` equivalent.

Logic around `deliverError` / `emitError` too. Let's do some shared utils here

packages/boltz-swaps/src/statusSource/websocket.ts
```diff
@@ -0,0 +1,452 @@
...
+    let ws: WebSocketLike | undefined;
+    let reconnectTimer: ReturnType<typeof setTimeout> | undefined;
+    let openTimer: ReturnType<typeof setTimeout> | undefined;
+    let pingTimer: ReturnType<typeof setInterval> | undefined;
+    let awaitingPong = false;
+    let attempt = 0;
+    let fallbackActive = false;
+    let sourceClosed = false;
+
+    const resolveConstructor = (): WebSocketConstructor | null => {
```
jackstar12: Can't we run this once on init and be done with it instead of multiple times?

packages/boltz-swaps/src/statusSource/websocket.ts
```diff
@@ -0,0 +1,444 @@
...
+        socket.onopen = (): void => {
+            if (ws !== socket) {
+                return;
+            }
+            clearOpenTimer();
+            attempt = 0;
+            deactivateFallback();
+            send("subscribe", Array.from(subscribers.keys()));
+            startPinging(socket);
+        };
```
kiocos, line 331: Instead of resetting `attempt` immediately `onopen`, could we reset `attempt` only after the socket has stayed open for a bit of time?

I'm afraid resetting immediately could causes a flood of requests if this ends in a loop where backend accepts the connection and drops right after for some reason.

PR #1550, opened 2026-07-06
feat: deposits in SDK

michael1011:


packages/boltz-swaps/src/deposit/types.ts
```diff
@@ -0,0 +1,225 @@
...
+    "USDC-ETH": 12,
+    "USDC-POL": 64,
+    "USDC-BASE": 12,
+};
+
+// The canonical asset every deposit bridges to before swapping out.
+export const DEPOSIT_BRIDGE_ASSET = "USDC";
+
+// Terminal chains a deposit can settle out to.
+export type DepositChainOut = "BTC" | "L-BTC";
```
kilrau, line 39: Plan to add RBTC in this PR?

michael1011: No

PR #1557, opened 2026-07-09
fix: blocked non-renegotiable hop lockup shortfalls

michael1011:
## Summary by CodeRabbit

* **New Features**
  * Added support for checking whether a lower-than-expected hop quote can still be locked for execution.
  * Improved quote handling so valid quotes can proceed even when they fall short of the initial threshold, when permitted.

* **Bug Fixes**
  * Prevented signing/locking transactions when a quote is below the required minimum and cannot be safely accepted.
  * Updated retry behavior to avoid unnecessary locking attempts for quotes that are not recoverable.

### DIFF
--- a/src/components/LockupEvm.tsx
+++ b/src/components/LockupEvm.tsx
@@ -1,5 +1,6 @@
 import type { RouterCall } from "boltz-swaps/bridge";
 import {
+    type Pairs,
     type QuoteData,
     encodeDexQuote,
     getCommitmentLockupDetails,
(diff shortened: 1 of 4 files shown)


src/components/LockupEvm.tsx
```diff
@@ -322,6 +326,47 @@ const lockupWithHops = async (
...
 
+        // The exact-input quote can fall below the swap's expected amount if the
+        // rate moved; block a non-renegotiable shortfall before signing rather
+        // than locking into an "insufficient amount" rejection.
+        if (hopInputAmount !== undefined) {
+            const quotedLockupAmount = BigInt(quote.quote);
+            if (
+                quotedLockupAmount <
+                calculateAmountOutMin(lockupAmount, slippage)
+            ) {
```
kiocos, line 337: For commitment swaps this will never be true, since `lockupAmount` would be 0 and the quoted amount wouldn't be less than 0.

Maybe we should have a separate `if` with validations for commitment swap amounts?

PR #1563, opened 2026-07-13
fix: racy 0 amount swap quote

kiocos:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved auto-claim reliability by using persisted swap quote/claim amounts and applying slippage thresholding from fresh quote state.
  * Added safer “invalid persisted claim state” handling (including closed-on-zero behavior) and improved retry on auto-claim failure.
  * Prevented non-positive amounts from triggering DEX quote requests.
  * Improved synchronization by serializing replacement-quote acceptance and ensuring claiming uses the locked, updated swap.

* **Tests**
  * Expanded coverage for persisted-amount logic, locking/concurrency behavior, retry flows, and reverse-swap claiming.


src/utils/chainSwapQuote.ts
```diff
@@ -0,0 +1,42 @@
+import BigNumber from "bignumber.js";
+
+// Same-tab serialization fallback for browsers without the Web Locks API.
+const fallbackQueues = new Map<string, Promise<unknown>>();
```
michael1011: We already require the Web Locks API. Don't reinvent it

src/context/Pay.tsx
```diff
@@ -253,10 +301,36 @@ const PayProvider = (props: { children: JSX.Element }) => {
...
+                    claimableSwap =
+                        claimableSwap.type !== SwapType.Chain
+                            ? null
+                            : await recoverChainSwapAmounts(
+                                  claimableSwap as ChainSwap,
+                                  transaction.hex,
+                              );
+                    if (claimableSwap === null) {
+                        log.warn(
+                            `Deferring claim of swap ${swapId}: receive amount is not positive yet`,
```
michael1011: What triggers the rerun of this?

src/context/Pay.tsx
```diff
@@ -253,10 +301,36 @@ const PayProvider = (props: { children: JSX.Element }) => {
                     return swaps;
                 });
 
+                const transaction = data.transaction as { hex: string };
+                let claimableSwap = await withChainSwapQuoteLock(swapId, () =>
```
michael1011: The lock only wraps the read, so an acceptance of a new quote could land between read and claim call. Unlikely, but might as well fix this bug for good

michael1011 (changes requested):

michael1011 (approved):

PR #1568, opened 2026-07-15
fix: allow direct EVM rescue claims without a wallet

kiocos:
## Summary by CodeRabbit

* **New Features**
  * ERC20 claims can now continue without a connected wallet when the claim destination can be safely determined.
  * Improved handling of claim destinations for direct claims and routed swaps.

* **Bug Fixes**
  * Prevented claims from proceeding when the destination cannot be recovered.
  * Updated claim instructions across English, German, Spanish, Portuguese, Chinese, and Japanese translations.


kilrau (dismissed):
Still doesn't look correct to me:

Image

Image

Image

michael1011 (approved):

